Issue #1550, opened 2026-04-25
EPIC: a1/1 fix — 17 issues blocking automatic pipeline pass + gpt-5.5 writer A/B


krisztiankoos:
## Why this exists

Tonight (2026-04-24/25) the convergence-loop fix #1545 merged successfully and the infrastructure is proven. But a1/1 still terminal-failed because the surrounding pipeline has 17 distinct bugs/conflicts that prevent the module from passing review even with the convergence fix in place.

User explicit: **no rerun until ALL 17 problems fixed.** Manual content patches do not count. The pipeline must produce passing modules unaided.

Full handoff doc: [`docs/session-state/2026-04-25-a1-1-fix-comprehensive-handoff.md`](https://github.com/learn-ukrainian/learn-ukrainian.github.io/blob/main/docs/session-state/2026-04-25-a1-1-fix-comprehensive-handoff.md)

### A. Pipeline / prompt bugs
- [ ] **A1** INLINE_MIN/MAX hardcoded — prose has 8 markers but prompt says max 6 (impossible math)
- [ ] **A2** Configs treated as MAX, but CLAUDE.md says they're MIN (project-wide sweep needed)
- [ ] **A3** Forbidden vs Recommended type list conflicts in `v6-activities.md`: `translate`, `error-correction`, `unjumble` get conflicting signals
- [ ] **A4** `{TYPE_DIVERSITY}` placeholder degrades to literal `0`, producing "MUST use at least 0 distinct activity types"
- [ ] **A5** Plan vocabulary doesn't include 33 letters or their key_words → letter-grid items fail grounding
- [ ] **A6** Seminar types (~80 lines) included in non-seminar prompts (pure distraction)
- [ ] **A7** Topic-specific patterns reference textbook §s the writer can't access (lines 538–630 of activity prompt)
- [ ] **A8** `exercise_quality` plan_revision_request triggered by activity ORDER not matching `activity_obligations` — should be set-membership, not order-strict

### B. Reviewer calibration (USER POINT — A1 M1-M3 special case)
- [ ] **B1** Dialogue dim has no awareness that A1 M1-M3 modules CAN'T have rich dialogue. Beginners haven't learned enough Ukrainian. Reviewer over-strictness, not content failure. **Fix**: per-module pedagogical context injection into reviewer prompts.

### C. Content issues (writer-prompt fixes)
[...]

PR #1556, opened 2026-04-25
fix(activities): resolve prompt contradictions + plan vocab coverage (#1550 U3)

krisztiankoos:
Refs #1550

- Reconciles activity prompt count/type instructions: inline count now derives from marker count, type diversity renders as a positive floor, and the checklist no longer permits self-defeating zero diversity.
- Adds required-vocab coverage, strict grounding, and letter-module exception blocks to the activity prompt.
- Guards seminar activity type reference so non-seminar prompts do not render the seminar-type block.
- Adds activity pre-validation before review for grounding and required vocabulary coverage, with one regeneration/repair retry before failing.
- Raises A1 audit min_types_unique to 4, matching the prompt floor needed by letter modules.
- Adds a config invariant test for forbidden vs allowed/priority activity type lists.
- Updates a1/1 with letter_module: true, bumps the plan version, records #1550 in plan_fixes, and adds all 33 Ukrainian letters with key_word examples.

Notes:
- The stale handoff items about hardcoded seminar lists and textbook § patterns were not present in the 352-line prompt; the PR only guards the actual seminar type reference.
- The brief's later constraints said not to touch a1/2 or a1/3, so this PR only changes a1/1.


krisztiankoos:
Fix-up pushed in a21033d068.

Summary:
- step_activity_pre_validate now skips cleanly when the activities artifact has no V6 activity items: list-shaped degraded artifacts, absent or empty YAML, or dicts with empty inline/workbook sections.
- Dict-shaped activities with actual items still run through verify_exercises; success-path behavior is unchanged.
- The skip writes an activity-pre-validation.json payload with reason: no activities to validate, and logs the skip.

Verification:
- PASS: /Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/pytest tests/test_v6_build_events.py -v -k "skipped_optional or degraded"
- PASS: /Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/ruff check scripts/build/ tests/
- Full pytest was attempted locally; it is not green in this worktree because broader tests depend on missing local resources like the worktree-local .venv/bin/python, data/vesum.db, embed-venv/bin/python, and wiki/source DB tables, plus existing unrelated suite expectation failures. The three PR #1556 regressions pass.

PR #1646, opened 2026-05-02
fix(v7_build): render authoring activity schema in writer prompt

krisztiankoos:
## Summary
- Render writer-facing activity schema from authoring fields consumed by scripts/yaml_activities.py instead of React component props.
- Clarify the v7 writer prompt so quiz/select/translate use items, not questions.
- Add regression coverage for rejecting quiz.questions and for prompt schema rendering.

## Verification
- .venv/bin/pytest tests/build/ -x
- .venv/bin/ruff check scripts/build/
- pre-commit: ruff + affected tests
- Real M20 rerun passed the writer phase; next blocker is downstream python_qg plan_sections after ADR-008 correction.

## Review
- Gemini cross-review: no blockers, verdict [AGREE].

Generated M20 artifacts were cleaned and are not included.

### DIFF
--- a/scripts/build/linear_pipeline.py
+++ b/scripts/build/linear_pipeline.py
@@ -691,94 +691,38 @@ def writer_context(plan: Mapping[str, Any], plan_content: str, knowledge_packet:
     }
 
 
-# Multi-line TSDoc comments embedded in raw type strings would leak into the
-# rendered writer prompt verbatim — `odd-one-out`'s `items` field, for example,
-# carries `{ /** * @schemaDescription Words... */ words: string[]; ... }[]`.
-# That burns tokens and confuses the writer with TypeScript syntax. Strip them
-# before rendering. Generator-side fix tracked in #1604.
-_TSDOC_BLOCK_RE = re.compile(r"/\*\*.*?\*/", re.DOTALL)
-
-
-def _strip_tsdoc(raw_type: str) -> str:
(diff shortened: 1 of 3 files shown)


krisztiankoos:
Gemini review requested by Codex/user; Claude was not used.

Verdict: [APPROVE]

Gemini summary:
- PR #1646 correctly fixes the v7 writer failure by exposing the authoring field whitelist instead of React component props.
- The fix addresses the quiz.questions vs quiz.items mismatch at the prompt/schema layer without weakening the strict parser.
- Prompt directives and regression tests are appropriate for preventing recurrence.

Findings:
1. Nit: _strip_tsdoc / _TSDOC_BLOCK_RE became dead code after moving away from lesson-schema TypeScript prop rendering. Addressed in amended commit ed4b923618 by removing the helper and obsolete test.
2. Minor follow-up: the new compact authoring schema no longer includes nested item field descriptions. This is non-blocking for M20, but a future improvement should enrich the authoring schema with nested item signatures so writers get guidance for item shapes such as match-up.left/right and true-false.statement/correct.

Local verification after addressing the nit:
- .venv/bin/python -m pytest tests/build/test_linear_pipeline.py -q: 73 passed
- .venv/bin/ruff check scripts/build/linear_pipeline.py tests/build/test_linear_pipeline.py: passed
- pre-commit on amended commit: passed

PR #1647, opened 2026-05-03
ci: scope workflow token permissions for CodeQL alerts

krisztiankoos:
## Summary
- Add explicit workflow-level permissions: contents read to CI, YAML validation, and rules deployment workflows.
- Add job-level actions write only for jobs that use dependency caching or upload artifacts.
- Addresses the low-risk CodeQL actions/missing-workflow-permissions alert slice in .github/workflows/ci.yml, .github/workflows/validate-yaml.yml, and .github/workflows/rules-deployment-check.yml.

## Verification
- .venv/bin/python YAML parse/assert check for the three touched workflows.
- git diff --check for the touched workflow files.
- Gemini review initially requested changes because cache/artifact jobs need actions scope; amended commit c21bb8ac1a adds job-level actions write only for those jobs.

## Notes
- Secret scanning API still reports secret scanning is disabled on this repository as of this check; I will retry after GitHub propagates the setting.
- This PR intentionally does not attempt to fix the remaining Python/JS CodeQL alerts; those need separate scoped PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+permissions:
+  contents: read
+
 on:
   push:
     paths:
@@ -239,6 +242,9 @@ jobs:
   lesson-schema-drift:
     name: Lesson Schema Drift
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: write
     needs: changes
     if: needs.changes.outputs.code == 'true'
     steps:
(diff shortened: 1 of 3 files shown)


krisztiankoos:
Gemini re-review requested by Codex/user; Claude was not used.

Verdict: [APPROVE]

Gemini previously flagged that contents-read-only could break cache/artifact jobs. The amended commit c21bb8ac1a keeps a restrictive top-level contents: read default and adds job-level actions: write only where caching or artifact upload is used.

Gemini confirmed the blocker is resolved and the PR now follows least privilege while preserving CI stability.

PR #1732, opened 2026-05-06
channels.html UX overhaul (Part A) + multi-UI participation ADR (Part B) (#1731)

krisztiankoos:
Closes #1731 Part A. Files Part B as PROPOSED ADR for cross-agent review.

## Part A — Surgical UX overhaul (`playgrounds/channels.html`)

The dashboard had five blockers preventing it from being useful during live `ab discuss` deliberations:

| # | Pain | Fix |
|---|---|---|
| 1 | Auto-refresh wiped innerHTML every 5s, destroying scroll + expand state | Delta-render: rendered-id Set, only NEW messages appended via `insertAdjacentHTML`, existing nodes preserved |
| 2 | Message bodies truncated to 300 chars (discussion rounds run 1500-2500) | Removed truncation entirely; `.messages-area` handles vertical scroll |
| 3 | No thread filter — architecture channel routinely has 10+ threads | Auto-shown thread dropdown (hides for <2 threads) with root-excerpt + max-round + participant labels |
| 4 | User posts blended in with agent posts | Orange left-border + 👤 prefix + tinted body background |
| 5 | No live-discussion indicator | Green-pulse strip with per-agent state (done ●, agreed ✓, pending ⏳) + "Pin to this thread →" jump link |

### Verification

Inline browser check via claude-in-chrome MCP:
- 14 threads in architecture channel dropdown ✓
- 12 user posts with `rgb(240,136,62)` (--orange) border ✓
- Longest message renders 8422 chars in full ✓
- Scroll position preserved across forced background refresh ✓
- Synthetic fresh-data activates strip with correct done/agreed states ✓
- Thread filter narrows 50→7→0→50 across selection cycles ✓

### Adversarial review

Gemini (`gemini-3.1-pro-preview`) returned REVISE with five findings, all incorporated:
- `escapeHTML` now escapes `'` (XSS defense-in-depth)
- `groupByThread` runs once per tick, passed to both refreshers
- Thread-bar dropdown skips innerHTML rewrite when option set unchanged (no popup-collapse mid-click)
- Discussion strip skips innerHTML rewrite when state unchanged
- `[AGREE]` regex case-insensitive but still end-anchored (no mid-message false positives)

Boy-Scout cleanup: dead `window.expandMessage` + `.expand-link` CSS removed.

## Part B — Multi-UI participation ADR (PROPOSED)

`docs/decisions/pending/2026-05-06-multi-ui-channel-participation.md` answers the six architectural questions from #1731 Part B with concrete proposals + tradeoffs:

1. **Participant** = `(agent_family, ui_surface, instance_id)` tuple, wire format `participant_id`.
2. **Discovery** — three-tier: SSE push, inbox-poll fallback, MCP tool shim for Claude surfaces.
3. **Arbitration** — claim per `(thread, round, agent_family)`, 90s expiry.
4. **Rounds** — user posts are NEVER replies, ALWAYS context; user `[AGREE]` is hard stop, not consensus.
5. **Auth** — localhost-only honour system; remote auth deferred.
6. **Failure** — 90s claim expiry + auto-fallback to legacy subprocess-spawn.

Closes with seven targeted review questions for Gemini + Codex.

**Per user mandate, zero Part B code ships until ACCEPTED.**

## Test plan

- [x] JS parses clean (`node --check` on extracted script block)
- [x] Browser smoke test (5 features verified via claude-in-chrome MCP)
- [x] Gemini adversarial review (REVISE → 5 fixes applied → re-verified)
- [ ] Codex adversarial review on Part B ADR (post-merge follow-up)
- [ ] User signoff to flip Part B from PROPOSED → ACCEPTED (post-merge follow-up)

### DIFF
--- a/docs/decisions/pending/2026-05-06-multi-ui-channel-participation.md
+++ b/docs/decisions/pending/2026-05-06-multi-ui-channel-participation.md
@@ -1,0 +1,270 @@
+# DECISION REQUIRED — Multi-UI agent participation in `ab discuss` / channels.html
+
+**Status:** PROPOSED (awaiting cross-agent review by Gemini + Codex, then user signoff)
+**Surfaced:** 2026-05-06 morning, while running the writer-lock 3-way discussion
+**Source:** GH issue #1731 Part B
+**Predecessor ADRs:** None — this is the first ADR on multi-UI participation. Touches the `ab` channel-bridge contract from #1190.
+
+---
+
+## Why this needs an ADR (not just a PR)
+
+`ab discuss` today is a **closed-loop, single-orchestrator** protocol. The agent that calls `ab discuss --with claude,gemini,codex` is a process that:
+
+1. Spawns each named agent as a subprocess (one per round, per agent) via `scripts/agent_runtime/adapters/*.py`.
+2. Reads each subprocess's stdout, posts the result to the channel, threads via `parent_id`.
+3. Aggregates `[AGREE]` markers; short-circuits when all participants agree, or runs to `--max-rounds`.
+4. Returns the transcript to the caller.
+
+Three properties hold today only because of this closed-loop assumption:
+
+- **Identity** is fixed at dispatch. `--with claude,gemini,codex` is the participant list; nothing else can join.
+- **Time** is owned by the orchestrator. A round is "Gemini's reply has returned from `gemini -p ...`" — there's no clock outside the subprocess.
+- **State** is a single transcript. The orchestrator is the only writer of `parent_id` chains; there is no claim/lock for "who responds next."
+
+The user's ask — "Claude Code, Claude Desktop, Codex CLI, Codex UI, AND user can all participate" — breaks ALL THREE properties:
+
+1. Claude Desktop is a separate process. It does not know `ab discuss` was invoked unless it polls or subscribes. → identity becomes dynamic.
+2. Claude Desktop's reply latency is unbounded (user might walk away). → time can no longer be owned by the orchestrator.
+3. If two Claude instances see the same channel post, both could reply. → state needs arbitration.
+
+Patching `ab discuss` ad hoc would silently change the contract for every existing caller (the writer-lock discussion, every channel review, every pipeline review). That's the rule-violation we want to avoid. Hence: ADR first.
+
+---
+
+## Scope of THIS ADR
+
+**In scope:** the participation/identity/round-semantics contract for an `ab discuss`-style multi-agent thread that can include Claude Code (CLI), Claude Desktop, Codex CLI, Codex UI, Gemini CLI, and human user.
+
+**Out of scope** (explicitly deferred to follow-up issues, even if the design touches them):
+
+- Real-time UI niceties (typing indicators, presence beyond "live now") — `playgrounds/channels.html` already has the basics, more polish is Part A territory.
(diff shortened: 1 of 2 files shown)


krisztiankoos:
[REVISE Q1: schema is internally inconsistent.] Lines 58-66 propose tuple identity but only add `participant_id`; lines 74-76 then claim family/surface/instance are “first-class” query dimensions. They are not first-class if two are buried in a parsed string. Store `agent_family`, `ui_surface`, `instance_id` as columns, with optional `participant_id` as canonical/display text or generated key. This avoids parser bugs, escaping/versioning issues, weak indexes, and painful future queries like “all live desktop Claude instances.” I would reject Q1 as written.

[REVISE Q2: discovery needs durable delivery semantics.] SSE + inbox + MCP is directionally right, but lines 94-98 underspecify replay/cursor behavior. `discussion_started` cannot be transient; a Desktop/UI client that connects 2 seconds late must still discover it. Require `event_id`/`since`/Last-Event-ID semantics and define whether MCP “subscribe” actually wakes a model or just exposes a readable resource. Without that, Desktop participation is aspirational.

[AGREE Q3] Family-scoped claims are correct for “tag in” semantics, and lines 135-143 make the right tradeoff. Counter-case: if a caller explicitly wants `claude:cli` and `claude:desktop` as independent viewpoints in the same quorum, family-scoped claims collapse them. That should be modeled as distinct quorum slots, not the default. Add an escape hatch like `participant_scope=family|participant` later; do not make instance-scoped the base contract.

[REVISE Q4: user-as-context default is right, but explicit `--with user` needs a defined mode.] Lines 160-167 are correct for ordinary mid-discussion steering. Steel-man: if the user explicitly runs `ab discuss --with claude,gemini,user`, they are asking to be a quorum participant, not just context. That is useful for design decisions where “human accepts/blocks” is part of convergence. I would keep default “user context + hard stop,” but add an explicit `user_quorum=true` / `--with user` semantic where user posts can satisfy the `user` slot, and user `[AGREE]` counts only for that slot, not as global hard stop unless marked terminate.

[AGREE Q5] Localhost honour-system is acceptable for this repo’s dev workflow, with one condition: line 204 must become a pre-implementation gate, not “separate fix.” If `:8765` binds beyond loopback, no-auth is no longer acceptable. Also log `client_app` as proposed in lines 191-193.

[REVISE Q6: 90s fixed expiry is too brittle.] Lines 214-222 mix LLM latency and human think-time. 90s is too short for Desktop/user-mediated takeover, but too long for a UI with no visible progress. Use a stateful claim: initial 90s lease, 30s keepalive extension, visible `expires_at`, and a manual “release/fallback now.” For claims made by automated clients, 90s is fine; for human/UI claims, allow a longer lease cap, e.g. 5 minutes with keepalive and clear UI countdown.

Missing architectural questions before implementation:

1. Round identity and idempotency: claim endpoint needs idempotency keys/client tokens so reconnect/retry does not create ambiguous holder state. Lines 128-137 do not specify this.
2. Message validation: posting a round reply must verify holder, `round_index`, `agent_family`, and parent/thread lineage atomically, not just rely on clients behaving.
3. Dynamic participant list: lines 98 and 124-125 imply invited participants, but there is no rule for joining after `discussion_started`: observer only, eligible claimant, or quorum member?
4. Event ordering: define monotonic sequence IDs across `message_appended`, `claim_created`, `claim_expired`, `discussion_aborted`; channels.html will otherwise show impossible states.
5. Fallback identity: line 217 says Desktop can degrade to CLI. The transcript must record both “required family=claude” and “actual participant=claude:cli:...” so later audit does not misattribute the reply.

PR #1796, opened 2026-05-07
fix(ab-bridge): preserve discuss root and infer ask sender

### LINKED ISSUE #1786: ab discuss + ask-* infrastructure bugs (root truncation, plan-mode tool subset, --from default)
## Background

Per `docs/session-state/2026-05-07-kubedojo-paradigm-followups.md` Action B. Three related infrastructure bugs surfaced by a failed `ab discuss` round on 2026-05-07. All three are independently confirmed by Codex's diagnosis (preserved in `architecture` thread `1c1b5d54966742ffacd1bf60e0893c1c`) and the fresh-Claude subagent's first-hand observation.

---

## B.1 — `ab discuss` truncates root message out of round-2+ prompts

**Symptom**: Round-2+ agents in any `ab discuss` operate on a brief that has the original question stripped out, especially in busy channels (architecture has 157+ prior messages → high truncation pressure).

**Diagnosis (verbatim from fresh-Claude subagent in failed discussion)**:
> "In round 2 my prompt shows `... [15 older messages omitted] ...` — the root message was truncated out entirely from my view. With `needed_history = 1 + N*max_rounds + 10` (see `scripts/ai_agent_bridge/_channels_cli.py:1256`), the design is to preserve the root, so the truncation happened at the prompt-render boundary, not from history sizing."

**Fix scope**: in `_channels_cli.py:1256` (or the prompt-render boundary nearby), pin the root message into the prompt unconditionally — it's the question, it must always be visible. Truncate against history, not the root.

**Test**: discussion with `--max-rounds 4` in a channel with 200+ prior msgs; assert root message appears in round-4 prompt verbatim.

---

## B.2 — Fresh Claude subprocess in `ab discuss` is instantiated in plan mode WITHOUT writer tools

**Symptom**: Round-1 and round-2 Claude replies in the failed discussion were both "I have a tooling problem" complaints — Claude correctly refused to fake responses, but couldn't engage with the brief.

**Available tools observed**: `Glob`, `Grep`, `LSP`, `Read`, MCP `sources`, MCP `claude_ai_Google_Drive`. Missing: `Write`, `Edit`, `Bash`, `Agent`, `ExitPlanMode`, `AskUserQuestion`.

**Two possible fixes** (investigate which is right):
1. Don't put discuss subagents in plan mode (they're commenting, not planning)
2. If plan mode is intentional, attach `Write` + `ExitPlanMode` so the subagent can complete the workflow

**Reference**: claude subagent's round-2 reply has full diagnosis with file refs (preserved in `architecture` thread `1c1b5d54966742ffacd1bf60e0893c1c`).

---

## B.3 — `ask-codex` / `ask-gemini` defaults `--from` to "gemini"

**Symptom**: When Claude (orchestrator) calls `ask-codex` without `--from claude`, the broker logs `From: gemini → To: codex`. Cosmetic but misleading for telemetry / channel attribution / future audit reports.

**Fix**: detect caller identity from environment (e.g. `CLAUDE_AGENT_NAME` env var, or absence of `CODEX_SESSION` env), or require explicit `--from` instead of defaulting to "gemini".

---

## Acceptance criteria

- B.1 fix lands with regression test (root preserved in round-N prompt for N up to max-rounds)
- B.2 fix decided + applied (option 1 or 2) with regression test asserting subagent has `Write` + `ExitPlanMode` (or that plan mode is off)
- B.3 fix lands with regression test asserting `--from` defaults to caller-detected identity, not hardcoded "gemini"
- All three covered by a single PR or three small PRs (Codex's call)

## Dispatch suggestion

Codex worktree dispatch. ~100-150 LOC across `scripts/ai_agent_bridge/_channels_cli.py` and the discuss subagent spawning code. Single PR preferred for atomicity (test fixtures share setup).

cc/related: #1779 (inbox TTL is the orthogonal architectural fix), #1782 (persistent listeners)

krisztiankoos:
## Summary

Closes #1786.

- Pins the original `ab discuss` root question into every round prompt outside truncatable channel history, so round 2+ and round 4 prompts always include the actual question even when older history is omitted.
- Fixes Claude discuss subagent spawning via Option A: discussion agents are commenting, not producing implementation plans, so Claude no longer runs with `--permission-mode plan`. It still gets the restricted `Read,Grep,Glob,LS` tool list and the bridge still fails the round if the worktree mutates.
- Changes legacy `ask-*` CLI sender defaults from hardcoded `gemini` to explicit caller detection from environment (`CLAUDE_AGENT_NAME`, `CODEX_SESSION`, Claude wrapper env, `GEMINI_SESSION`) and fails loudly when no identity can be inferred.
- Updates `docs/best-practices/agent-bridge.md` for the visible Claude discuss and `ask-* --from` behavior.

## B.2 Decision

Option A is the right fix. The adapter and docs showed Claude plan mode was intentional as a safety mechanism, but `ab discuss` participants are not being asked to draft or exit an implementation plan. The plan-mode workflow itself caused the missing-tool complaints (`Write` / `ExitPlanMode`). Keeping Claude out of plan mode while restricting built-in tools to read/list/search preserves the discussion contract without requiring plan-mode completion tools.

## Tests

- `.venv/bin/ruff check scripts/ai_agent_bridge/_cli.py scripts/ai_agent_bridge/_gemini.py scripts/ai_agent_bridge/_channels_cli.py scripts/agent_runtime/adapters/claude.py tests/ai_agent_bridge/test_ab_discuss_bugs.py tests/test_agent_runtime.py tests/test_gemini_bridge.py`
- `.venv/bin/pytest tests/ai_agent_bridge/ -x`
- `.venv/bin/pytest tests/test_agent_runtime.py::test_claude_adapter_discussion_readonly_uses_restricted_tools_without_plan_mode tests/test_gemini_bridge.py::test_handle_ask_gemini_passes_auth_mode -q`
- Commit hook: ruff plus affected pytest suite, 132 passed

Note: full `.venv/bin/ruff check` still reports pre-existing unrelated issues in `docs/reports/*` and `plans/*`; this PR does not modify those files.

### DIFF
--- a/docs/best-practices/agent-bridge.md
+++ b/docs/best-practices/agent-bridge.md
@@ -62,8 +62,10 @@ Current adapter policy for discussion calls:
 - Codex: `read-only` maps to `codex exec -s read-only`.
 - Gemini: discussion calls add `--approval-mode plan`; plain read-only
   CLI defaults were not enough in trusted workspaces.
-- Claude: discussion calls add `--permission-mode plan` and restrict
-  built-in tools to read/list/search tools.
+- Claude: discussion calls restrict built-in tools to read/list/search
+  tools. They do not use Claude plan mode because discussion replies are
+  comments, not implementation plans; the restricted tool list plus the
+  bridge's git mutation guard enforce the read-only contract.
 
 ## CLI quick reference
 
@@ -261,8 +263,10 @@ to 40–60KB of copy-pasted transcript.
 one-time cost — the pinned `context.md` + the Monitor API snapshot
 are concatenated into the prompt even if the agent doesn't strictly
 need them. For drive-by one-shots, the legacy `ask-*` commands are
-still cheaper. Use channels when the conversation will have at
-least two turns.
+still cheaper. `ask-*` commands infer `--from` from wrapper environment
+such as `CLAUDE_AGENT_NAME`, `CODEX_SESSION`, or `GEMINI_SESSION`; pass
+`--from` explicitly when running them from a plain shell. Use channels
+when the conversation will have at least two turns.
 
 ## Web dashboard
 
--- a/scripts/agent_runtime/adapters/claude.py
+++ b/scripts/agent_runtime/adapters/claude.py
@@ -273,10 +273,7 @@ class ClaudeAdapter:
         cmd.append("--verbose")
 
         if discussion_readonly:
-            cmd.extend([
-                "--permission-mode", "plan",
-                "--tools", "Read,Grep,Glob,LS",
-            ])
+            cmd.extend(["--tools", "Read,Grep,Glob,LS"])
 
         # Mode-specific flags
         if mode == "danger":
--- a/scripts/ai_agent_bridge/_channels_cli.py
+++ b/scripts/ai_agent_bridge/_channels_cli.py
@@ -60,6 +60,21 @@ _GEMINI_AUTH_CHOICES = ["auto", "subscription", "api-key", "api"]
 _DISCUSSION_READONLY_TOOL_CONFIG = {"discussion_readonly": True}
 
 
+def _build_discuss_round_body(root_body: str, directive: str) -> str:
+    """Render the high-priority body for a discuss round prompt.
+
+    The root question is pinned outside channel history because history
+    truncation is allowed to drop old messages. In later rounds, verbose
+    replies can exceed the history budget before the root is rendered.
+    """
+    return (
+        "--- discussion root question (pinned) ---\n"
+        f"{root_body}\n\n"
+        "--- round directive ---\n"
+        f"{directive}"
+    )
+
+
 def _deadline_seconds_arg(raw_value: str) -> int:
     try:
         seconds = int(raw_value)
@@ -1279,7 +1294,7 @@ def _handle_discuss(args) -> int:
         try:
             prompt_obj = _channels.build_agent_prompt(
                 args.channel,
-                directive,
+                _build_discuss_round_body(body, directive),
                 history_tail=needed_history,
                 review=args.review,
             )
--- a/scripts/ai_agent_bridge/_cli.py
+++ b/scripts/ai_agent_bridge/_cli.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import os
 import sys
 from datetime import UTC, datetime
 from pathlib import Path
@@ -35,6 +36,46 @@ from ._messaging import (
     send_message,
 )
 from ._model import check_model
+
+_CALLER_IDENTITY_ENV_HINTS = (
+    "CLAUDE_AGENT_NAME",
+    "CODEX_SESSION",
+    "CLAUDE_PROJECT_DIR",
+    "CLAUDE_CODE_FILE_READ_MAX_OUTPUT_TOKENS",
+    "GEMINI_SESSION",
+)
+
+
+def _detect_caller_identity_from_env() -> str | None:
+    """Infer the sending agent for legacy ask-* commands from wrapper env."""
+    claude_name = os.environ.get("CLAUDE_AGENT_NAME")
+    if claude_name:
+        return claude_name.strip().lower()
+    if os.environ.get("CODEX_SESSION"):
+        return "codex"
+    if (
+        os.environ.get("CLAUDE_PROJECT_DIR")
+        or os.environ.get("CLAUDE_CODE_FILE_READ_MAX_OUTPUT_TOKENS")
+    ):
+        return "claude"
+    if os.environ.get("GEMINI_SESSION"):
+        return "gemini"
+    return None
+
+
+def _resolve_from_llm(args) -> str:
+    """Return explicit --from or fail loudly if caller identity is unknown."""
+    explicit = getattr(args, "from_llm", None)
+    if explicit:
+        return explicit
+    detected = _detect_caller_identity_from_env()
+    if detected:
+        return detected
+    hint_list = ", ".join(_CALLER_IDENTITY_ENV_HINTS)
+    raise SystemExit(
+        "Cannot infer sender for ask-* command. Pass --from explicitly or "
+        f"run under a known agent wrapper that sets one of: {hint_list}."
+    )
 
 
 def interactive_mode():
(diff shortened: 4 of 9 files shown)


krisztiankoos:
# Adversarial review by Claude headless (xhigh)

**Verdict**: REVISE
**Confidence**: high

I reviewed the diff, ran the targeted bridge tests locally (10/10 pass), traced the prompt-render path in `_channels_cli.py`/`_channels.py`, and probed the three sub-bugs against their actual mechanism. The fixes themselves are well-architected. However, **CI is red** — the signature change in `_send_gemini_message` was not propagated to existing callers, so `tests/test_coverage_misc.py::TestSendGeminiMessage` fails three tests. That's blocking.

## Strengths
- **B.1** correctly identifies the root cause: `directive` was the only thing pinned outside the truncatable history, so root contents (channel post #1) was at the mercy of `truncate_history_by_budget`. Wrapping `body + directive` into the new "body" passed to `build_agent_prompt` puts the root in the unbreakable section. Clean helper `_build_discuss_round_body` keeps the intent legible.
- **B.2** picks the right option (A — drop plan mode) and explains why in the PR body. Discussion rounds are commenting, not implementation planning, so plan mode was misclassified. The remaining `--tools Read,Grep,Glob,LS` whitelist + `_discussion_worktree_changed` post-hoc guard preserve the read-only contract.
- **B.3** fails loud (`SystemExit`) when caller identity can't be inferred — strictly better than silently mislabeling as "gemini". `--from` precedence is sensible: explicit > `CLAUDE_AGENT_NAME` > `CODEX_SESSION` > Claude wrapper-env proxies > `GEMINI_SESSION`.
- The new test `test_discuss_round_four_prompt_pins_root_after_history_truncation` is a faithful repro: 205 prior msgs + 4×6KB fake replies forces real truncation (`DEFAULT_MAX_HISTORY_CHARS=6000`), then asserts both `... [` truncation marker AND `root_body` are present in round-4 prompt.

## Issues found

### 1. [BLOCKING] `tests/test_coverage_misc.py::TestSendGeminiMessage` — 3 tests broken by signature change
- **File:** `tests/test_coverage_misc.py:436, 444, 452`
- **Symptom:** `TypeError: _send_gemini_message() missing 1 required positional argument: 'output_path'`
- **Cause:** The diff (PR line 192) widened `_send_gemini_message`'s signature from 8 to 9 positional params by inserting `from_llm`:
  ```python
  -def _send_gemini_message(content, task_id, msg_type, data, from_model, model, ...
  +def _send_gemini_message(content, task_id, msg_type, data, from_llm, from_model, model, ...
  ```
  But the existing test's positional call still passes 8 args, so every arg shifts by one and `output_path` is missing.
- **Reproduce locally:** `cd .worktrees/dispatch/codex/1786-ab-discuss-bugs && /Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/pytest tests/test_coverage_misc.py::TestSendGeminiMessage -x`
  → 1 failed in 0.16s (and the other 2 fail with the same error).
- **CI also confirms:** `Test (pytest)` failure on the open PR — `actions/runs/25527415468/job/74926152506`.
- **Fix:** insert the new `from_llm` arg in the three test calls, e.g.:
  ```python
  result = f("content", "task-1", "query", None, None, None, "model", False, "/out")
  ```
  or refactor to kwargs.
- **Sibling check:** `Grep`d for other callers of `_send_gemini_message` — only the bridge calls it internally and these tests poke it via reflection. Just the 3 sites need updating.

### 2. [IMPORTANT] B.1 fix introduces behavioral regression for huge root briefs
- **File:** `scripts/ai_agent_bridge/_channels_cli.py:1294-1300`
- Previously only `directive` (~1 KB) was treated as "body" by `build_agent_prompt`. Now `_build_discuss_round_body(body, directive)` pins both — meaning root briefs above ~22 KB will exceed `max_prompt_chars=24000` (set in `_channels.py:135`) and `build_agent_prompt` raises `ValueError`. The `try/except ValueError` at line 1301 catches it and the discussion exits 1.
- Real-world impact: any user pasting a long brief (e.g., 30 KB session-state file as `--body`) will see "❌ prompt build failed" where previously the round would just have history-truncated more aggressively.
- Mitigation options: (a) raise `max_prompt_chars` for discuss flows (it's a tunable kwarg), (b) accept the regression and document it in `agent-bridge.md`, (c) detect oversize root and warn-and-truncate the body itself with a clear marker.

### 3. [IMPORTANT] Caller-env detection uses a fragile proxy
- **File:** `scripts/ai_agent_bridge/_cli.py:42-58`
- The Claude-detection proxies are `CLAUDE_PROJECT_DIR` (stable, harness-set) and `CLAUDE_CODE_FILE_READ_MAX_OUTPUT_TOKENS` (internal Claude-CLI tunable that could be renamed at any version bump).
- `CLAUDE_PROJECT_DIR` alone is enough; drop `CLAUDE_CODE_FILE_READ_MAX_OUTPUT_TOKENS` to avoid coupling to internal CLI knobs.
- Document in `agent-bridge.md` that wrappers SHOULD set `CLAUDE_AGENT_NAME=claude` explicitly — the env-proxy ladder is a fallback, not the primary signal.

### 4. [NIT] `_handle_ask_gemini` switch is duplicated across both branches
- **File:** `scripts/ai_agent_bridge/_gemini.py:163-184`
- Both `if output_path:` and `else:` branches re-do `if from_llm: send_message(...) else: send_to_gemini(...)`. Could be hoisted into a single helper that returns the `msg_id` then the branch only handles the acknowledge/print difference.

### 5. [NIT] B.3 is a hard breaking change with no soft-warn period
- Scripts that previously omitted `--from` and relied on the default "gemini" silently now hit `SystemExit`. This is the intended behavior per the brief, but worth a CHANGELOG / `agent-bridge.md` note so dashboard/cron callers can be retrofitted before they break.

## AC coverage check
- **B.1**: ✓ root preserved through round-4 prompt under 200+ msg busy-channel pressure (test verifies)
- **B.2**: ✓ Option A applied; `--permission-mode plan` removed; `--tools Read,Grep,Glob,LS` retained; rationale in PR body
- **B.3**: ✓ caller-env detection ladder + loud-fail on unknown identity; tests cover both detection-success and unknown-fail paths
- Single PR atomicity: ✓
- `docs/best-practices/agent-bridge.md` updated: ✓
- `ruff check` clean on touched files: ✓ (verified)
- `.venv/bin/pytest tests/ai_agent_bridge/ -x` passes: ✓ (10/10 locally)
- **Broader pytest passes**: ✗ — `tests/test_coverage_misc.py::TestSendGeminiMessage` (×3) fails locally and in CI

## Recommendation
**REVISE.** Issue #1 is blocking — CI is red and the failure is a direct consequence of this PR's signature change. Fix is mechanical (insert one positional arg in 3 test calls). Issues #2 and #3 are worth addressing in the same revision; #4-5 are follow-up-grade.

After issue #1 is fixed and CI is green, this is mergeable. The architectural choices are sound.

[OBJECT]

PR #5114, opened 2026-07-14
fix(lexicon): explicit preserve-vs-retract gate semantics for offline enrich (#5077)

krisztiankoos:
## Problem (tool-verified)

`LEXICON_SLOVNYK_OFFLINE=1` enrich on the hydrated 8,552 release stripped **809 sections to empty** and **shrank 1,748** vs the published asset. Root cause traced in `enrich_manifest.py::enrich_entry`: it recomputes gated sections from scratch each run and unconditionally does `entry["sections"] = sections`. A gate that **cannot run** (offline; per-lemma slovnyk cache reset on a `schema_version` bump then never refetched; transient error) returns empty and silently overwrites the previously-confirmed published section. `verify_manifest`'s NONEMPTY→EMPTY hazard caught only the empties, not the partial shrinks.

## Design — three-outcome contract at the gate layer (root cause, not per-callsite)

`enrich_entry` now captures the hydrated/published sections as an explicit **preserve baseline** (threaded through the function, no hidden global) and resolves each gated section via `_resolve_gated_section`:

| Outcome | Condition | Action | `gate_provenance` |
|---|---|---|---|
| ran-and-confirmed | gate ran, no confirmed item lost | take new (additions welcome) | *(none — default, minimal diff)* |
| ran-and-rejected | gate ran, dropped item(s) | take new (quality win) | `rejected` |
| did-not-run | gate could not consult its source **and** would drop confirmed item(s) | **preserve existing byte-for-byte** | `skipped-offline` |

**"Did the slovnyk gate run" is honest, not offline-blind** (`_slovnyk_gate_ran`): online always ran; offline ran **iff** the per-lemma cache already holds lookups — an empty `lookups` map is exactly the schema-reset/never-fetched lemma the offline gate cannot consult. So an offline run over already-cached lemmas still applies authoritative retractions; only genuinely-unconsulted lemmas are preserved. Paronyms are local ZNO/cache only → their gate always runs (`gate_ran=True`).

**Design point 4 honored:** when the stricter gate *can* run, its retractions (WordNet auto-translation junk `ключ→джерело/живець`, `адвокат→прихильник/радник`, `аеродром→поле`) are `gate_ran=True` → taken as authoritative and marked `rejected`, never blocked.

## verify_manifest shrink gate

`verify_manifest.py --baseline <hydrated>` adds a **per-section item-count shrink gate**. A regression fails the promote **unless** justified by:
- the entry's `gate_provenance[section] == rejected` (the gate ran — a quality win), **or**
- the `(lemma, section)` pair on the curated `scripts/lexicon/shrink_allowlist.yaml` (explicit intended-retraction mechanism).

This is the second line of defence the empty-section hazard lacked — it also catches partial shrinks.

## Verification evidence (tool-backed, #M-4)

**Affected + wider suites — 159 in the commit gate, all green:**
```
tests/test_lexicon_enrich_manifest.py ... tests/test_verify_manifest.py ... tests/test_manifest_io.py
============================= 159 passed in 1.24s ==============================
```
Wider sweep (conformance, publish, perf, reenrich, grow_*, fill_local, wiki, manifest_api, fingerprint, calque): all green (`96 passed, 1 skipped`; `29 passed`; `62 passed`).

**New tests (design step 4):**
- `test_offline_gate_did_not_run_preserves_section_byte_identical` — (a) preserve byte-identical + `skipped-offline`.
- `test_online_gate_ran_and_rejected_retracts_with_provenance` / `..._partial_retraction_keeps_survivors_marks_rejected` — (b) retraction allowed + `rejected`.
- `test_offline_with_cached_lookups_counts_as_gate_ran` — cache, not offline, is the signal.
- `test_online_gate_confirmed_addition_records_no_provenance` — minimal-diff default.
- `test_shrink_gate_catches_unexplained_regression` / `_flags_vanished_section` / `_allows_allowlisted_retraction` / `_allows_gate_ran_rejection_via_provenance` / `_ignores_growth_and_equal` / `_off_without_baseline` — (c) shrink gate + both exemption paths.

**Real CLI (not just `run()`):**
```
# unexplained shrink (стіл, not allowlisted, no provenance)
=== SHRINK GATE (per-section item count vs baseline, #5077) ===
  стіл             synonyms     3 -> 1
=== VERDICT: 1 unexplained section shrink(s) — do NOT commit ===
exit code = 2
# allowlisted lemma (ключ, on shrink_allowlist.yaml) → CLEAN, rc=0
```

**Ruff:** `All checks passed!` on all changed files.

## Notes / decisions that deviate or need a reviewer eye

1. **Preserve is whole-section, not an item-level union.** The issue's design point 1 says a gate that did not run must *preserve the existing section*; the item-level merge is named only as the **one-time recovery**. I implemented the literal contract: on a did-not-run loss the baseline section is kept unchanged (byte-identical), and offline additions from still-available sources are **deferred to the next online run** (auditable via `gate_provenance: skipped-offline`). This is the conservative, non-destructive choice; flagging it explicitly in case a reviewer prefers the union for offline relation-only republishes.
2. **`gate_provenance` is a new top-level entry key**, recorded only for `rejected`/`skipped-offline` (never `confirmed`) to keep full-run diffs minimal. No lexicon-entry JSON schema constrains entry keys; conformance/hazard gates ignore it.
3. **`site/src/data/lexicon-manifest.fingerprint.json` is included** — it is the required `check_manifest_freshness` companion to the two changed code files (`manifest_io.py` is excluded from the fingerprint glob, so only the enrich/verify hashes moved). Freshness gate re-run: `rc=0`.
4. Shared `GATE_*` constants live in `manifest_io.py` (stdlib-only) so enrich and verify share them without either importing the other's heavy deps.

Refs #5077

### DIFF
--- a/scripts/lexicon/README.md
+++ b/scripts/lexicon/README.md
@@ -52,24 +52,38 @@ authorized. See `docs/bug-autopsies/atlas-manifest-source-split.md` for the
 
 ### Offline enrich is non-destructive (preserve-vs-retract, #5077)
 
-`enrich_manifest.py` recomputes gated sections (synonyms/antonyms/homonyms/idioms)
-from scratch each run. Some gates need a live lookup (the Ukrajinet WordNet gate
-wants СУМ-20/slovnyk confirmation; the per-lemma slovnyk cache is reset when its
-`schema_version` bumps). With `LEXICON_SLOVNYK_OFFLINE=1` — or after a cache reset,
-or a transient error — a gate that **cannot run** used to return empty and silently
-overwrite the previously-confirmed published section (809 sections stripped + 1,748
-shrunk on the 2026-07-13 go-live). Enrich now distinguishes three per-section
-outcomes and records the notable ones in an entry's `gate_provenance` map:
+`enrich_manifest.py` recomputes gated sections (synonyms/antonyms/homonyms/paronyms/
+idioms) from scratch each run. The **synonyms** and **idioms** sections need a live
+slovnyk lookup (synonyms from the slovnyk synonym dictionaries; idioms from the
+Фразеологічний page). With `LEXICON_SLOVNYK_OFFLINE=1` — or after the per-lemma cache
+is reset when its `schema_version` bumps, or a transient error — a gate that **cannot
+run** used to return empty and silently overwrite the previously-confirmed published
+section (809 sections stripped + 1,748 shrunk on the 2026-07-13 go-live). Enrich now
+distinguishes three per-section outcomes and records the notable ones in an entry's
+`gate_provenance` map:
 
 - **ran-and-confirmed** — gate ran, kept/added items (default; not recorded).
 - **ran-and-rejected** — gate ran and dropped items (a quality win, e.g. WordNet
   auto-translation junk `ключ→джерело/живець`); recorded as `rejected`.
-- **did-not-run** — gate could not consult its source; the existing section is
-  preserved byte-for-byte; recorded as `skipped-offline`.
+- **did-not-run** — gate could not consult its source; a confirmed item that would be
+  dropped is preserved byte-for-byte, and the non-consultation is always recorded as
+  `skipped-offline` (even when nothing was lost), so the audit trail distinguishes a
+  freshly-gated section from one carried over unconsulted.
 
-Because an offline gate that did not run **preserves** rather than recomputes, an
-offline run cannot add brand-new items to a preserved section — run online (the full
-recipe above, on a data-enabled checkout) to pick up new slovnyk-sourced chips.
+"Did the gate run" is decided **per relevant slug set**, not `bool(lookups)`: the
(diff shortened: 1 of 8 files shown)


krisztiankoos:
Delta verified by orchestrator: ran the branch's 8 new regression tests in a clean worktree at ee408730ac — all pass, covering the review's attack paths (partial-slug cache preserves + skipped-offline, local sections update offline, provenance recorded without loss, byte-identical pure-preserve, shrink gate flags skipped-offline drops as unexplained, pre-existing provenance semantics). Antonym transparency note assessed: slovnyk cache contributes only secondary annotations to antonyms, never item membership (checked _antonyms_wiktionary on the branch) — count-based shrink gate still guards; annotation carry-over filed as a minor follow-up. Review gate satisfied (grok-build full adversarial review + verified delta). Auto-merge enabled.

X-Agent: claude-atlas

PR #5318, opened 2026-07-16
fix(atlas): align exporter tokenization + practice resolver with Sqlite contract (Sol F001/F002/F007)

krisztiankoos:
## Summary

Fixes Sol review findings **F001 / F002 / F007** (atlas runtime-shard contract) so `scripts/atlas/export_runtime_shards.py` matches live `SqliteAtlasDataSource` semantics.

### F001 — component tokenization diverged from TS (`\p{L}\p{M}` only)
**Failure scenario:** Python `COMPONENT_TOKEN_RE` used `\w`, so digits became component chips. Real corpus: `мені-20-років` exported `{text:"20"}`; Sqlite does not. Stress-marked forms (combining U+0301) also risked splitting if Marks were dropped.

**Fix:** `find_component_tokens()` mirrors `/[\p{L}\p{M}]+(?:['’][\p{L}\p{M}]+)*/gu` via `unicodedata` categories (no new `regex` dependency). Shared vectors at `scripts/atlas/component_tokenization_vectors.json`, asserted from pytest + vitest.

### F002 — practice-level resolver missed reconciliation
**Failure scenario:** Exporter indexed both `lemmaId` and `lemma` from a single `--deck-dir`. Sqlite prefers `api/lexicon/` before `lexicon/` and indexes **`lemmaId` only** (lookup still falls back to `entry.lemma`).

**Fix:** `practice_index_dir_candidates()` + `load_practice_levels_by_slug()` mirror Sqlite exactly. Hermetic disagreeing fixtures under `tests/fixtures/atlas/practice_decks/`.

### F007 — script path failed without `-m`
**Failure scenario:** `.venv/bin/python scripts/atlas/export_runtime_shards.py` → `ModuleNotFoundError: scripts`.

**Fix:** Bootstrap repo root onto `sys.path` when `__package__` is empty; subprocess `--help` test covers it.

### TOKEN_RE sibling check (no scope expansion)
py `TOKEN_RE` (`[\w\u0400-\u04ff]+`) vs `search.ts` `/[\p{L}\p{N}_]+/gu`: both include letters, digits, and underscore for search gloss tokens. No real divergence for Atlas UA/Latin glosses (edge: Python `\w` can match extra Unicode connectors beyond `_`; not observed on this corpus). Left a one-line cross-reference comment only.

### pytest green
```
$ .venv/bin/python -m pytest tests/test_export_runtime_shards.py -q
============================= 13 passed in 31.36s ==============================
```

### ruff clean
```
$ .venv/bin/ruff check scripts/atlas/ tests/test_export_runtime_shards.py tests/fixtures/atlas/
All checks passed!
```

### exporter still deterministic (fixture DB, twice)
```
A atlas-v1-af8a9ff7c313e91d
B atlas-v1-af8a9ff7c313e91d
diff_exit=0
```
(`diff -r` of the two out-dirs empty; dataVersion strings identical.)

### F001 fixed on real data (`--db data/atlas.db --no-decks`)
```
slug= мені-20-років
lemma= Мені 20 років
entry_type= multiword_term
componentLinks= [{"text": "Мені", "targetSlug": "я"}, {"text": "років", "targetSlug": "рік"}]
has_text_20= False
texts= ['Мені', 'років']
```

### TS/py vectors agree
```
$ cd site && npx vitest run tests/unit/atlas-runtime-shards.test.ts
 Test Files  1 passed (1)
      Tests  2 passed | 6 skipped (8)
```
(normalization + component tokenization vector suites; remaining tests skip without a prebuilt `build/atlas-runtime` tree.)

### F007 script-path smoke
```
$ .venv/bin/python scripts/atlas/export_runtime_shards.py --help
script_path_ok=1
```

## Test plan
- [x] pytest `tests/test_export_runtime_shards.py`
- [x] ruff on touched paths
- [x] double export + `diff -r`
- [x] real-DB probe for `мені-20-років`
- [x] vitest shared vectors
- [ ] CI on this PR


krisztiankoos (commented):
Cross-family review (Claude orchestrator, atlas lane; cursor-authored) — **REVISE, substance correct** (delta dispatched on this branch).

Verified myself before this verdict:
- **F001 real-data check PASSES**: exported real `atlas.db` on this head → `мені-20-років` componentLinks = `[Мені→я, років→рік]`, no `"20"` chip. Python `find_component_tokens` semantics traced against the TS regex incl. combining-mark and apostrophe-boundary edges — correct.
- **F002**: resolver semantics match `loadPracticeLevelsBySlug` on the parity-relevant paths (api-first per level, lemmaId-only, per-level fall-through); the new pytest pins dirs-disagree + lemma≠lemmaId. Accepted deviation, for the record: on a MALFORMED index file TS silently falls through to the next candidate, the exporter fails loud — fail-closed is right for a build tool.
- Ran the suite on this head in a clean worktree: **12 passed, 1 FAILED** — see finding 1.

Delta items (all small, exact):
1. **`test_script_path_invocation_exits_zero` fails in any added worktree** — it hardcodes `ROOT/.venv/bin/python`, which only exists in the primary checkout and CI. Every dispatch worker runs tests in a worktree; this is the green-in-CI/red-for-workers class. Fix: `sys.executable` (existing precedent in 6+ tests, e.g. `tests/test_repo_root.py:102`).
2. **The vitest vectors test asserts a TEST-LOCAL copy of `COMPONENT_TOKEN_RE`**, so drift in the production regex (`sqlite-atlas-data-source.ts:41`) would pass unseen — a gate that can pass while broken. Export the production constant (or a `findComponentTokens` helper) and assert THAT against the shared vectors.
3. **No numeric-multiword ENTRY exists in the fixture DB** — the tokenizer is vector-pinned but not end-to-end record-pinned. Add one (e.g. `мені-20-років` shape) via `build_runtime_shards_fixture.py`, regenerate the fixture DB, and assert the exported record's componentLinks carries no digit chip. This also feeds the upcoming unconditional parity gate (F006).

krisztiankoos (commented):
Delta verified (review follow-up) — **APPROVE** (comment-form: shared git identity).

All three findings fixed and re-verified by my own runs on head 2daa3c9b5e:
1. `sys.executable` in the script-path test → **13/13 pass in a venv-less added worktree** (previously 12+1 fail there) — my run, raw tail: `13 passed in 49.96s`.
2. `COMPONENT_TOKEN_RE` now exported from `sqlite-atlas-data-source.ts:42` and the vitest vectors test imports the PRODUCTION constant — regex drift now fails the gate.
3. Fixture DB regenerated (282→327 KB) with a numeric-multiword entry; new pytest asserts the exported record's componentLinks carries no digit chip (line 158).

Earlier real-data verification stands: exporting real atlas.db on this branch gives `мені-20-років` → `[Мені→я, років→рік]`, matching the Sqlite source. Merging on green CI.

PR #5319, opened 2026-07-16
fix(atlas): browser-portable HttpAtlasDataSource (Sol F003/F004/F005)

krisztiankoos:
## Summary

Fixes three **CONFIRMED** Sol review findings in `HttpAtlasDataSource` (merged in PR #5282). Design target change: dictionary shards are **GitHub Release assets fetched on demand by the browser** — no edge worker, no Node in production. The reader must be browser-portable (F004 is required, not cosmetic).

### F003 — pointer/manifest cached forever → stale after release rollover (P1)

**Failure scenario:** A long-lived client instance memoized `current.json` + manifest unconditionally. After a new release flips the pointer to `v2`, the same instance kept serving `v1` forever (stale shards / wrong `dataVersion`).

**Fix:** Re-resolve `current.json` behind a short TTL (default **60s**, injectable). Reload the manifest when `dataVersion` changes. Within one logical request (`getEntry` / `search` / `getDeck`), pin a `PinnedAtlasSession` (`manifest` + `versionDir`) end-to-end so a mid-request rollover cannot mix v1/v2 shard URLs. Existing `version_mismatch` retry-once guard retained.

### F004 — static `node:*` imports → not browser-loadable (P1, required)

**Failure scenario:** Static `node:zlib` / `node:fs` / `node:path` imports made the module unloadable in the browser bundle for client-rendered word pages.

**Fix:**
- Core `http-atlas-data-source.ts`: **zero** `node:*` imports. Gzip via `DecompressionStream("gzip")` + pluggable `decompress` hook.
- New Node-only `http-atlas-node.ts`: `createFileAtlasFetch`, `nodeGunzip`, `createNodeHttpAtlasDataSource`.
- npm script `check:atlas-browser-bundle` for repeatable esbuild browser-bundle verification.
- Tests import Node helpers from `http-atlas-node` (no re-export of Node code from core).

### F005 — search drops gloss-only hits (P1)

**Failure scenario:** Article admission only allowed lemma exact/prefix or roman prefix. The exporter indexes glosses and `rankSearchResults` scores gloss hits — HTTP search silently lost gloss-only results that `SqliteAtlasDataSource` returns.

**Fix:** `admitsSearchArticle` admits lemma/roman exact·prefix·substring **and** gloss substring before ranking (mirrors ranker surfaces Sqlite uses).

## Files

| File | Change |
| --- | --- |
| `site/src/lib/lexicon/http-atlas-data-source.ts` | TTL pointer resolve, session pin, gloss admission, browser gzip |
| `site/src/lib/lexicon/http-atlas-node.ts` | **new** Node file fetch + gunzipSync wiring |
| `site/tests/unit/atlas-runtime-shards.test.ts` | hermetic F003/F005 tests; Node import path |
| `site/package.json` | `check:atlas-browser-bundle` script |

### vitest green (incl. new tests)

```text
$ cd site && npx vitest run tests/unit/atlas-runtime-shards.test.ts

 RUN  v4.1.9
 ✓ tests/unit/atlas-runtime-shards.test.ts (11 tests | 6 skipped) 63ms

 Test Files  1 passed (1)
      Tests  5 passed | 6 skipped (11)
```

Unconditional new tests:
- F003: pointer rollover v1→v2 on reused instance + concurrent in-flight pin
- F005 unit: `admitsSearchArticle` gloss-only
- F005 parity: gloss query `trustworthy` → identical Sqlite vs Http over fixture export
- F004 source guard: core has no static `node:` imports

Existing real-DB/export `test.skipIf` gating left unchanged (F006 follow-up).

### type-safe (`npx tsc --noEmit`)

```text
$ cd site && npx tsc --noEmit 2>&1 | rg "src/lib/lexicon/http-atlas" || echo "PASS: zero tsc errors in http-atlas modules"
PASS: zero tsc errors in http-atlas modules
```

Repo-wide `tsc` still reports pre-existing errors outside this change (activity-kit react types, unrelated unit tests). **No errors** in `http-atlas-data-source.ts` / `http-atlas-node.ts`.

### browser-portable (F004)

```text
$ cd site && npx esbuild src/lib/lexicon/http-atlas-data-source.ts --bundle --platform=browser --external:astro* --outfile=/dev/null

  ../../../../../../../../../dev/null  18.6kb

⚡ Done in 9ms
```

```text
$ cd site && npm run check:atlas-browser-bundle

  ../../../../../../../../../dev/null  18.6kb

⚡ Done in 2ms
```

(Contrast: bundling `http-atlas-node.ts` for browser correctly fails on `node:fs` / `node:path` / `node:zlib` — Node-only by design.)

### rollover works (F003)

New vitest (always on): stub fetch serves v1 → exercise → flip `current.json` to v2 + advance clock past TTL → **same instance** returns v2 data. Concurrent in-flight request gated on v2 shard while pointer flips to missing v3 → finishes on v2 only; no v3 URLs fetched.

### gloss search parity (F005)

New vitest (always on): exports `tests/fixtures/atlas/runtime_shards_fixture.db` to a temp tree, isolates Sqlite to that DB (hides dual-publication search JSON for the test duration), query `"trustworthy"` (gloss-only for `достовірний`) — Sqlite and Http return the same result set + order; hit is not lemma/roman admitted.

## Out of scope

- F006 (`test.skipIf` gating of real-DB parity tests)
- Merge / auto-merge / self-review (orchestrator review gate)

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -20,7 +20,8 @@
     "test": "npm run hydrate && vitest run",
     "test:watch": "npm run hydrate && vitest --watch",
     "test:coverage": "npm run hydrate && vitest run --coverage",
-    "test:atlas-route-parity": "node --experimental-strip-types ./scripts/atlas-lexicon-route-parity.mjs"
+    "test:atlas-route-parity": "node --experimental-strip-types ./scripts/atlas-lexicon-route-parity.mjs",
+    "check:atlas-browser-bundle": "esbuild src/lib/lexicon/http-atlas-data-source.ts --bundle --platform=browser --external:astro* --outfile=/dev/null"
   },
   "dependencies": {
     "@astrojs/check": "^0.9.9",
--- a/site/src/lib/lexicon/http-atlas-data-source.ts
+++ b/site/src/lib/lexicon/http-atlas-data-source.ts
@@ -1,13 +1,12 @@
 /**
- * HTTP/file Atlas data source — reads an exported runtime shard tree.
+ * HTTP Atlas data source — browser-portable reader for exported runtime shards.
  *
- * Used for parity tests against SqliteAtlasDataSource and as the browser/VPS
- * reader shape. R2AtlasDataSource (Worker binding + Cache API) is a later PR.
+ * Production serves dictionary shards as GitHub Release assets fetched on demand
+ * by the browser (client-rendered word pages). This module must stay free of
+ * `node:*` imports. Node-only helpers (file fetch, gunzipSync) live in
+ * `http-atlas-node.ts`.
  */
 
-import { gunzipSync } from "node:zlib";
-import { readFileSync } from "node:fs";
-import { resolve } from "node:path";
 import {
   AtlasDataSourceError,
   isValidAtlasSlug,
@@ -26,6 +25,25 @@ import type { PracticeDeckData } from "./srs.ts";
 export interface AtlasFetch {
   (url: string): Promise<Uint8Array>;
 }
+
+/** Pluggable gzip decode — browsers use DecompressionStream; Node injects gunzipSync. */
+export type AtlasDecompress = (compressed: Uint8Array) => Promise<Uint8Array> | Uint8Array;
+
+export interface HttpAtlasDataSourceOptions {
+  assetBaseUrl?: string;
+  /**
+   * Max age for cached `current.json` / manifest pointer resolution.
+   * Default 60s. Set to 0 to re-resolve on every logical request (tests).
+   */
+  pointerTtlMs?: number;
+  /** Override gzip decode. Defaults to browser `DecompressionStream("gzip")`. */
+  decompress?: AtlasDecompress;
+  /** Clock for pointer TTL; injectable for tests. */
+  now?: () => number;
+}
+
+/** Default pointer TTL — short enough to observe release rollovers, long enough to amortize fetches. */
+export const DEFAULT_POINTER_TTL_MS = 60_000;
 
 interface CurrentPointer {
   schema: string;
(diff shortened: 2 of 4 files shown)


krisztiankoos (commented):
Cross-family review (Claude orchestrator, atlas lane; grok-authored) — **APPROVE** (comment-form: shared git identity blocks formal approval).

Independently verified, not just read:
- **F004**: `grep 'from "node:'` on the core module = 0 hits; ran the browser-bundle check myself: `npx esbuild <core>.ts --bundle --platform=browser` → clean 18.7 kb bundle. Node helpers correctly isolated in `http-atlas-node.ts`.
- **F003**: TTL'd pointer (60s default, injectable clock), manifest reload on dataVersion change, per-request pinned `{pointer, manifest, versionDir}` — mid-request rollover cannot tear shard URLs. Asset cache keyed `version:id:sha256` with byte-length + sha256 integrity checks retained — no stale-bytes path across rollover.
- **F005**: `admitsSearchArticle` is a strict superset of `matchTier`'s scoring surfaces, and the ALIAS admission got the same substring fix (`aliasMatchTier` tier-3 sibling — good catch, same bug class). Over-admission is parity-safe: `rankSearchResults` makes the final call, exactly like the Sqlite source which pre-filters nothing.

Scope note for the record (pre-existing shard design, NOT this PR's defect): mid-word substring hits homed in shards the query prefix doesn't select are structurally unreachable for the HTTP source (Sqlite full-scans; selection is a prefix trie). Rows are multi-homed by lemma/roman/gloss first-token chars, so token-initial queries have full parity — the F006 unconditional parity gate will use token-initial queries and document this boundary.

Merging on green CI (auto-merge armed).

PR #5809, opened 2026-07-25
fix(ci): green main — cache validation and replay reports

krisztiankoos:
## Summary
- share the committed source-inventory index across all 51 decision ledgers
- preserve the lifecycle fail-closed assertions while replaying the shadow report once

## Validation
- `.venv/bin/python -m pytest tests/test_source_inventory_review_decisions.py tests/orchestration/test_curriculum_lifecycle_pilot.py -q` → 36 passed (twice)
- `.venv/bin/ruff check scripts/audit/source_inventory_review_decisions.py tests/test_source_inventory_review_decisions.py tests/orchestration/test_curriculum_lifecycle_pilot.py`
- `.venv/bin/python scripts/audit/lint_opsec_leaks.py --staged`

No auto-merge requested.

### DIFF
--- a/scripts/audit/source_inventory_review_decisions.py
+++ b/scripts/audit/source_inventory_review_decisions.py
@@ -90,7 +90,15 @@ def validate_committed_decision_files(
     if not decision_paths:
         raise SourceInventoryError("no source-inventory review decision files found")
 
-    summaries = [validate_decision_file(path) for path in decision_paths]
+    # Every committed decision ledger is checked against the same immutable
+    # inventory corpus. Loading that corpus per ledger was needlessly expensive
+    # (and pushed the complete CI shard past pytest's per-test timeout). Keep one
+    # shared index for this validation pass; `validate_decision_file` extends it
+    # only when a ledger references a present staged inventory.
+    source_index = _source_record_index(
+        read_source_inventories(COMMITTED_SOURCE_INVENTORIES, project_root=PROJECT_ROOT)
+    )
+    summaries = [validate_decision_file(path, source_index=source_index) for path in decision_paths]
     decision_counts: Counter[str] = Counter()
     total_rows = 0
     for summary in summaries:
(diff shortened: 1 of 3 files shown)


krisztiankoos:
## The actual review body — recovered from the bridge; it was NOT published to this PR

The 145-character verdict comment above is the output of `publish-review-verdict`, which posts
only the verdict line plus provenance. **The reviewer's analysis was dropped by that publishing
step, not omitted by the reviewer.** Restoring it here so this gate is actually auditable.

- reviewer: `claude-sonnet-5` (anthropic family; author is codex/GPT, so cross-family holds)
- schema `code-review-findings.v1` · confidence **0.78** · correctness **correct** · findings **0**

### Analysis as written by the reviewer

> The refactor introduces a shared, incrementally-populated source_index across all decision files validated in one validate_committed_decision_files() call, replacing per-file re-reads of the immutable committed inventory corpus. The `raw_path not in indexed_inventory_paths` gate correctly skips re-reading any inventory whose records are already fully loaded (since read_source_inventories loads a whole file's records at once, indexed_inventory_paths always reflects complete coverage for any path it contains), and dict.fromkeys still dedups reads within a single file. The public single-file API (validate_decision_file/_index_and_absent_inventories with source_index=None) is unchanged in behavior, and resolve_staged_inventory_path's path-containment check still runs unconditionally per row regardless of cache hits, so the traversal safety boundary is preserved. The absent-inventory fail-open path (used for regenerable, gitignored staged corpora) is unaffected. The lifecycle test change replaces a second expensive build_shadow_report() call with deepcopy(valid_report) purely to avoid a redundant rebuild; deepcopy yields an independent object equivalent to a second call (backed by the existing reproducibility assertion in test_shadow_report_is_exactly_reproducible_at_one_source_tree), so this reduces replay cost without weakening any assertion, exception matching, or mutation isolation between the two tamper scenarios. The new test_committed_decision_validation_reuses_shared_source_index directly pins the intended single-read caching behavior. No validation gap, security regression, or test weakening was found.

---

### Why this matters beyond this PR

As published, this review was indistinguishable from a rubber stamp — I said so on this PR and
in #5802, and I was wrong about the reviewer. The substance existed the entire time and simply
never reached the PR.

A verdict published without its evidence turns a real review into an unfalsifiable assertion,
and it makes a diligent reviewer look negligent. It also actively defeats the mitigation I had
proposed in #5802 ("confirm the gate from the PR itself"), because the PR is precisely where
the evidence is missing. Filed against the bridge publishing path.

PR #6541, opened 2026-08-09
docs(hooks): document durable thread lease recovery

krisztiankoos:
## Summary

- replace the stale runbook statement that lease recovery is still in progress with the exact behavior already landed on `main`
- document PID/start-time/machine liveness, immediate reclaim of confirmed-dead or PID-reused owners, and fail-closed handling for uncheckable owners
- document the identity-fenced `SessionEnd` release tombstone and safe `SIGKILL` recovery posture

The executable fix is already present on `main` through #5758, #5845, and #5896. This PR closes the remaining operator-documentation and issue-linkage gap without duplicating the lease implementation.

## Design

The owner process identity is authoritative: confirmed-live owners are never stolen from, confirmed-dead owners are reclaimed immediately, cooperative release is identity-fenced, and an uncheckable owner requires the exact CAS-scoped operator release instead of a clock-based takeover.

## Validation

- focused lease claim/release/restart and ABA-fencing pytest: 9 passed
- `ruff check scripts/orchestration/thread_handoff.py tests/orchestration/test_thread_handoff.py`: passed
- Markdown lint on `docs/runbooks/codex-hooks.md`: passed
- `git diff --check`: passed
- agent-trailer and OPSEC audits: passed

A broader run of `tests/orchestration/test_thread_handoff.py` reached 148 passed and four unrelated subprocess tests could not execute because they hard-code a worktree-local `.venv/bin/python`; this dispatch worktree intentionally has no `.venv` and all project commands used the shared primary interpreter.

## Residual

When process liveness is genuinely uncheckable, the lease remains fail-closed until an operator runs the exact CAS-scoped release printed by the conflict. This is intentional to preserve anti-double-drive safety.

Fixes #5759

### DIFF
--- a/docs/runbooks/codex-hooks.md
+++ b/docs/runbooks/codex-hooks.md
@@ -175,10 +175,20 @@ accumulate past the outer hook deadline. Hook-owned advisory locks use a
 one-second bounded wait; the lock remains fail-safe, while a live but wedged
 owner can no longer stall a new session indefinitely.
 
-A `SIGKILL` can land after `claim-thread-lease` persists the lease but before the
-claim reports back to SessionStart. In that case the session must stop and the
-lease can remain held; do not force-release based only on a claim timeout. The
-lease-lifecycle work in the infrastructure lane is making this recovery explicit.
+A durable thread lease records the harness process PID, process start time, and
+machine identity. A new session reclaims the lease immediately when that exact
+process is dead or the PID has been reused; it never steals from a confirmed-live
+owner, regardless of heartbeat age. `heartbeat_at` is diagnostic only. If process
+liveness cannot be checked, the claim fails closed and prints a CAS-scoped
+`release-thread-lease --force` command bound to the observed owner and generation.
+Use only that exact command after verifying the owner is gone; a confirmed-live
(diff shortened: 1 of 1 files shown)


krisztiankoos (commented):
Cross-family review of exact head `9d0dfa70575a36082b93a169612b512f460d72ff` by Kimi K3 via the direct ACP review lane.

VERDICT: APPROVED

Verified against the code at this head:

- PID, process-start-time, and machine identity are persisted and used for owner liveness.
- Confirmed-dead, zombie, and PID-reused owners are reclaimed without a clock gate.
- Confirmed-live and uncheckable owners fail closed; heartbeat age is diagnostic only.
- Operator force release is CAS-scoped to owner plus generation, and live owners require explicit acknowledgement.
- SessionEnd calls the identity-fenced release path, writes a tombstone, and the next claim advances the monotonic generation.
- Claim timeout/ambiguous gate outcomes stop without being mislabeled as permission to force-release.

No blocking findings.

Informational only: this material lives in the Codex hooks runbook while the durable lease is currently claimed only for Claude-family lanes. The guidance describes the lease system accurately and does not create an unsafe operator action; a lane-scoping note could be a later clarity improvement.

PR #6828, opened 2026-08-15
feat(lexicon): make grow_lexicon_from_content crash-safe and resumable (#5214)

### LINKED ISSUE #5214: [atlas][grow] grow_lexicon_from_content: no checkpoint + silent-for-hours makes long runs fragile & unobservable
## Problem
`scripts/lexicon/grow_lexicon_from_content.py` runs a rate-limited per-lemma enrichment loop (Wikimedia/slovnyk fetches) that can take **11+ hours** on the full content delta, but:

1. **No checkpoint / single terminal write.** `grow_candidates.json` is written exactly once, at the end (`DEFAULT_OUT.write_text(...)`). A crash/reboot/kill at hour 11 loses the entire run — nothing is salvageable.
2. **Fetches are not persisted across runs.** `_preserve_wiki_reference_cache()` snapshots the wiki-reference cache at start and **reverts it in `finally`**, so a restart re-fetches everything from scratch. The run is not cheaply resumable.
3. **Silent by design.** stdout is empty until the final write, so there is no progress signal, no ETA, and no way to distinguish "healthy but slow" from "hung" without external forensics (CPU-time deltas, `lsof` sockets, `sources.db-wal` growth).

Observed 2026-07-15: a live grow run (pid 2354) had run 11h+ with an empty log; confirming it was progressing required `command ps` CPU-time sampling + `lsof` socket inspection. That's the wrong diagnostic surface for a load-bearing pipeline step.

## Proposed fix (root cause)
- **Incremental checkpointing:** flush a partial `grow_candidates.json` (or `.partial` sidecar) every N lemmas so a crash preserves work; resume from the last checkpoint.
- **Persist enrichment cache across runs** (or make the revert opt-in) so a restart hits cache instead of re-fetching — turns an 11h re-run into minutes.
- **Progress logging:** emit a JSONL heartbeat (`processed/total`, ETA) on an interval, so the standard `Monitor` tool can watch it instead of `lsof`.

## Related
- Handoff "ARCHITECTURE LESSON" already flagged the worker-lifetime version (detached run = correct); this issue is the script's own resumability/observability design.

krisztiankoos:
## Summary
Closes #5214.

Previously, 11+ hour runs of `grow_lexicon_from_content` lost all accumulated progress if interrupted (single terminal write), and fetch caches were reverted on exit in `_preserve_wiki_reference_cache()`, forcing restarts to re-fetch from zero.

This PR makes candidate growth durable and resumable:
1. **Periodic atomic checkpointing**: Intermediate candidate payloads are saved every `--checkpoint-interval` (default 10) lemmas using atomic tempfile replace (`_write_atomically`), so a crash during write never corrupts previous checkpoints.
2. **Crash-safe resume**: On startup, `grow_lexicon_from_content` loads existing checkpointed entries from `--out` and reuses them without re-calling `enrich_entry` for already-processed lemmas.
3. **Persistent fetch caching**: Removes the `finally` cache reversion so newly fetched Wikipedia, Slovnyk, and GRAC cache entries persist on disk across runs.
4. **Progress visibility**: Added informative stderr progress logging with heartbeat updates every checkpoint interval.
5. **Rejected alternative**: Considered using an append-only JSONL sidecar / SQLite checkpoint database, but rejected it because it introduces secondary file drift, requires an extra compaction step, and complicates downstream consumers (`promote_grow_candidates.py`, `triage_needs_review.py`, CI workflows) which already expect the canonical JSON schema at `--out`. By writing the canonical payload atomically at each interval, `--out` remains 100% valid and self-contained at any point during or after a run.

## Verification
- Kill-resume test: verified on a multi-lemma fixture that an interrupted run saves checkpointed entries and the resumed run skips already-processed lemmas.
- Atomic write test: verified that simulated write failures do not corrupt existing checkpoint files.
- Mutation checks: verified that disabling checkpoint loading or using direct non-atomic writes fails guard tests.
- All test suites green: `pytest tests/test_grow_lexicon_from_content.py tests/test_grow_lexicon_from_sources.py tests/test_grow_lexicon_from_lemmas.py tests/test_promote_grow_candidates.py`
- Linter clean: `ruff check scripts/lexicon/grow_lexicon_from_content.py tests/test_grow_lexicon_from_content.py`

### DIFF
--- a/scripts/lexicon/grow_lexicon_from_content.py
+++ b/scripts/lexicon/grow_lexicon_from_content.py
@@ -9,12 +9,14 @@ Run from the repository root:
 from __future__ import annotations
 
 import argparse
-import copy
+import contextlib
 import json
+import os
 import sqlite3
 import sys
+import tempfile
 from collections.abc import Iterator, Sequence
-from contextlib import contextmanager
+from contextlib import contextmanager, suppress
 from pathlib import Path
 from typing import Any
 
@@ -29,6 +31,7 @@ from scripts.lexicon.content_lexicon_reconciler import (
 from scripts.lexicon.lemma_normalization import strip_acute_stress
 
 DEFAULT_OUT = PROJECT_ROOT / "data" / "lexicon" / "grow_candidates.json"
+DEFAULT_CHECKPOINT_INTERVAL = 10
 GENERATED_FROM = "content_lexicon_reconciler.missing_lemmas"
 _WARNING_CLASSIFICATIONS = {"russianism", "sovietism", "surzhyk"}
 _POS_PRIORITY = {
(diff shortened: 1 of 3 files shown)


krisztiankoos:

Reviewer: Claude Fable (inline orchestrator seat; cross-family vs author agy/Google), head 1a6907063d.

Verified by direct probe: (1) `_write_atomically` is the correct same-dir tmp + fsync + os.replace pattern with failure cleanup, and the corruption-on-failure test pins it; (2) kill-resume test proves completed work survives and already-enriched lemmas are not re-fetched; (3) the root-cause fix is real — `_preserve_wiki_reference_cache` no longer reverts caches in `finally` (the re-fetch-from-zero bug), and both flush targets exist in enrich_manifest (verified at :5487/:6419); (4) ran the full test file myself: 13 passed; ruff clean; (5) interval=0 edge is guarded on both the modulo and logging paths; negative interval rejected at the parser.

Non-blocking notes for the lane (no action required this PR): `_flush_caches` suppresses all exceptions silently — a persistently broken cache path would degrade to re-fetching without a trace; a one-line stderr warning would keep it observable. And a completed prior output file acts as a resume source for later runs by design — the payload's processed/total_delta counts make partialness visible, which is acceptable.

Merging on green CI.

Issue #6818, opened 2026-08-14
[infra][runtime] agy dispatch harness rewrote the PRIMARY checkout's node_modules as symlinks into its worktree — took down every ask transport


krisztiankoos:
## What happened (tool-backed timeline, 2026-08-14)

At 18:44 local, during the agy dispatch `atlas-6801-round2-panel` (which then died with a generic runtime error), the agy harness ran a pnpm operation that REWROTE the primary checkout's `node_modules`: 24 top-level entries (typescript, puppeteer, prettier, ts-jest, c8, markdownlint-cli2, `@agentclientprotocol/*`, …) became symlinks into `.worktrees/dispatch/agy/atlas-6801-meaning-choice-teach/node_modules/.pnpm/...`, and the project-local `acpx` package was clobbered (`node_modules/acpx/dist/cli.js` gone, dangling `.bin` shim left).

Consequence: every ACP ask adapter (codex/kimi/glm) failed its binary capability probe from that moment — the mid-session reviewer-transport collapse documented on #6805 (\"missing capabilities: root help, exec\"). The masked secondary finding: after pulling main with the glm-5.3 contract (#6800), the probe surfaces `--no-fs`, requiring acpx ≥0.13 per package.json — but the underlying breakage was the missing package, not the contract.

## Recovery performed (atlas driver)

- Removed the 24 foreign symlinks; full `rm -rf node_modules && npm ci` at primary root.
- npm's strict allow-scripts needed approvals (fsevents, better-sqlite3, protobufjs, puppeteer, unrs-resolver, esbuild).
- better-sqlite3 12.5.0 does NOT compile under local Node 26.7 (V8 `PropertyCallbackInfo::This` removed) and has no 26.x prebuilds — rebuilt under Homebrew node@22 (CI-matched). Local default `node` being 26.x is a latent footgun for anyone reinstalling.
- Verified restored: acpx 0.13.0 with `--no-fs`; `ask-glm` returns a real reply (`OK glm-5.3`).

## Asks (infra/runtime lane)

1. Root-cause how the agy worktree's pnpm run escaped into the primary's node_modules (shared store config? `node_modules` symlink-through in the sparse worktree? pnpm workspace resolution walking up?) and make dispatch worktrees incapable of writing the primary's node_modules.
[...]

Issue #6805, opened 2026-08-14
[infra][comms] Reviewer-seat transport attrition: hermes binary missing, ask-glm returns no findings, ask-gemma unrouted, grok 402


krisztiankoos:
## What

During one atlas drive (2026-08-14), four consecutive cross-family reviewer routes failed before a working one was found, all tool-backed:

1. **ask-hermes (DeepSeek code-review seat, #M0 default):** `AcpxDeepSeekShadowAdapter: hermes binary not found on PATH; required for compatibility contract 'text-oneshot-isolated-v1'` — the designated CODE-review lane is unreachable on this host.
2. **ask-glm:** transport reports `outcome=ok`, but the only captured output is the model's preamble line ("I'll verify key claims…") — no verdict, no findings ever delivered. The ACP event store is body-free by design, so the reply is unrecoverable; transport-ok masks an empty outcome.
3. **ask-gemma:** `legacy ask target 'gemma' has no enabled ACP route` — the $0 toolless seat (ideal for inline-diff review) is unrouted under authority mode.
4. **grok lane:** `402 Payment Required — Grok Build usage balance exhausted` (also recorded on #6803).

Working: ask-codex (delivered a real advisory verdict same session).

## Why it matters

Reviewer-seat economics route the bulk of reviews to non-Claude lanes, and the review gate is load-bearing for every merge. With deepseek/glm/gemma/grok all degraded, reviews funnel into codex/claude — exactly the concentration the routing rules exist to avoid. Worse, failure mode 2 is silent: a driver that doesn't read review CONTENT could count transport-ok as a completed review.

## Asks (infra/comms lane)

- Restore or formally retire the hermes binary path for the DeepSeek seat on this host.
- ask-glm: capture and deliver the full reply body (or fail loudly when the reply is empty) — outcome=ok with zero findings must not be representable.
- Decide gemma's ACP routing status under authority mode; the ask target should either work or be removed from the bridge target list.
- Surface per-lane reachability (binary present, route enabled, balance non-zero) in `fleet_comms plane-status` or `capacity_pick` so drivers see dead seats before routing to them (capacity_pick currently shows all lanes `unknown` on empty budget snapshots).

X-Agent: claude-atlas

krisztiankoos:
[...]

PR #6829, opened 2026-08-15
feat(delegate): node_modules integrity probe + Node-version pin (#6818)

krisztiankoos:
## Scope: Option B detection ONLY — ALERT-only, no behavior change

Implements the approved slice of #6818 per the design gate (original note,
K3 cross-family review REQUEST_CHANGES, revised note round 2 adopting K3's
root-cause correction). This PR is **detection only**: it never blocks a
dispatch and never auto-repairs anything.

### What landed

1. **Node-version pin** (`.nvmrc` + `package.json` `engines: node 22.x`,
   matching CI's `node-version: '22'`) — the hard precondition K3 required
   before any future auto-repair work (repairing under the wrong Node risks
   reproducing the better-sqlite3 native-module rebuild failure documented
   in the #6818 incident log). No auto-repair in this PR.
2. **`scripts/audit/check_node_modules_integrity.py`** — a
   `check_node_modules_integrity`-shaped probe mirroring
   `check_primary_integrity.py`'s detect-record-never-repair posture, scoped
   to the actual conduit K3 identified: `_provision_data_symlinks`
   symlinks the primary's `node_modules`/`site/node_modules` directly into
   every dispatch worktree, so a worktree write lands straight in the
   primary's real files (#6805 incident).
   - Detects symlinks under `node_modules/`/`site/node_modules/` (depth ≤2,
     including `.bin`) whose **realpath** resolves outside the primary
     root — not `*worktrees*` string matching, so macOS's `/var` vs
     `/private/var` aliasing and normal relative `.bin` shims never
     false-positive.
   - Sentinel check: `node_modules/acpx/dist/cli.js` exists and
     `.bin/acpx` resolves (the shared ACPX install every seat depends on).
   - Records a JSONL event (task_id/agent/pid attribution of in-flight
     dispatches) mirroring the watchdog's RECORD contract.
   - Fails open on any probe error, same contract as the existing watchdog.
3. Wired **ALERT-only** into the same three call sites
   `check_primary_integrity` already occupies: the pre-dispatch gate and
   `_run_worker` post-exit sweep in `scripts/delegate.py`, and the
   Monitor API health-orient canary in `scripts/api/main.py`.
4. `scripts/api/preload.py` registers the new module for the repo's
   import-pinning guard.

### Explicit non-goals (operator-gated, per the reviewed design)

- Does **not** touch `_provision_data_symlinks`.
- Does **not** block dispatch scheduling on detection.
- Does **not** implement auto-repair.
- Provisioning-level prevention (read-only symlink enforcement) stays on
  the operator decision list.

### Verification

- New tests: `tests/test_check_node_modules_integrity.py` (12 cases —
  outside-resolving symlink detected, `.bin` shim rewired to a worktree
  path detected [the exact #6805 incident shape], healthy relative `.bin`
  shims and scoped-package symlinks NOT flagged, depth >2 out of scope,
  missing/dangling sentinel detected, attribution recorded, fail-open
  contract, CLI exit codes).
- Mutation-checked the core detection function (`find_outside_symlinks`):
  reverting it to a no-op fails 5 of the 12 tests as expected; restored and
  green.
- Zero-baseline proof against the real primary post-`npm ci` state:
  ```
  $ .venv/bin/python scripts/audit/check_node_modules_integrity.py --repo /Users/krisztiankoos/projects/learn-ukrainian
  ✅ node_modules integrity ok (/Users/krisztiankoos/projects/learn-ukrainian)
  ```
  (confirms real `.bin` shims resolve inside the primary and are not
  false-flagged.)
- `.venv/bin/ruff check` clean on all changed/new Python files.
- `.venv/bin/pytest` green: 327 passed across
  `test_check_node_modules_integrity.py`, `test_check_primary_integrity.py`,
  `test_check_self_symlinks.py`, `test_delegate_primary_integrity.py`,
  `test_orient_api.py`, `test_delegate.py` (242 cases), and
  `test_import_pinning.py`.

Refs #6818.

X-Agent: claude-sonnet


krisztiankoos:

Reviewer: Kimi K3 (task review-6829-r4; the same seat that authored the design-review constraints), head 70e907d385. All 7 constraints verified PASS: ALERT-only invariant (no probe result feeds control flow; hard gate remains primary-integrity only), realpath-based outside-primary detection (macOS /var aliasing cancels; depth ≤2 + .bin with per-shape regressions; no path-string matching), sentinel correctness (dangling-shim shape tested), attribution via the watchdog's own _append_event/_running_dispatches (identical by construction, proven end-to-end), fail-open at all three call sites, zero-baseline + revert-failing tests (reviewer ran 12/12 green itself), _provision_data_symlinks untouched. Four non-blocking nits recorded in the review transcript (N1–N4) — none require changes; N2 (root-itself-a-symlink not walked) is worth a comment in a future touch.

Operator-gated remainder tracked on #6818: auto-block flip, auto-repair enablement, provisioning-level prevention. Merging on green CI.

krisztiankoos:
## Root cause — CI shard 4/4 red on `test_run_worker_periodic_stdout_avoids_silence_timeout`

Reproduced locally first: the test passed every time on this machine (5/5 runs), which
pointed at an environment-dependent probe rather than a real timing race.

**Cause:** this PR's `_run_worker` post-exit sweep calls
`check_node_modules_integrity()` (`scripts/audit/check_node_modules_integrity.py`),
which walks real symlinks under `delegate._REPO_ROOT` and checks a real `acpx`
sentinel (`node_modules/acpx/dist/cli.js` / `.bin/acpx`). It mirrors the existing
`check_primary_integrity` post-worker sweep in every way — including that the
existing sweep already has a dedicated autouse test stub
(`_stub_primary_integrity_sweep` in `tests/test_delegate.py`) precisely because its
verdict depends on the ambient checkout. The new node_modules sweep never got the
matching stub.

On a detached CI checkout (`actions/checkout`) that hasn't provisioned the
project-local acpx install, the sentinel check fails → `ok=False` →
`_append_dispatch_event("node_modules_integrity_post_worker", ...)` fires and creates
`dispatch_events.jsonl` under the test's `tmp_tasks_dir`. The test asserts
`not (tmp_tasks_dir / "dispatch_events.jsonl").exists()`, so it flips to
`assert not True` on that host. Locally the ambient checkout's `node_modules` is
intact, so the probe returns `ok=True`, no event is appended, and the assertion
never fires — hence green on main/locally, red on this shard.

## Fix

Added `_stub_node_modules_integrity_sweep`, an autouse fixture in
`tests/test_delegate.py` mirroring `_stub_primary_integrity_sweep`: monkeypatches
`check_node_modules_integrity` to a fixed `(True, ...)` verdict so `_run_worker`/
`cmd_dispatch` tests stay hermetic from the ambient checkout's real node_modules
state. Does not touch the probe's own behavior — it stays ALERT-only/fail-open in
`scripts/delegate.py`, and its real-fixture coverage
(`tests/test_check_node_modules_integrity.py`) is unchanged.

## Verification

- `pytest tests/test_delegate.py -q` → **242 passed**
- `pytest tests/test_check_node_modules_integrity.py tests/test_check_primary_integrity.py tests/test_delegate_primary_integrity.py -q` → **46 passed**
- `ruff check tests/test_delegate.py scripts/delegate.py scripts/audit/check_node_modules_integrity.py` → clean

Pushed to `claude/infra-6818-detection-impl` at `3ac655779e`. Auto-merge is already
armed — no further action needed unless CI surfaces something new.

X-Agent: claude-sonnet

PR #6832, opened 2026-08-15
fix(review): normalize combining stress marks in vocabulary surface candidates (#5375)

### LINKED ISSUE #5375: Normalize combining stress marks in vocabulary surface candidates
Parent stream epic: #4274

A1 special-signs currently resolves many stressed learner surfaces to zero vocabulary candidates because combining acute marks make exact/VESUM lookup miss otherwise valid forms (for example м'який vs м'яки́й). This can create false MISSING findings in canonical post-build review.

Acceptance: normalize combining stress marks only for candidate matching while preserving displayed evidence; verify the full a1/special-signs vocabulary ledger and add regressions.

krisztiankoos:
Closes #5375.

## What

The vocabulary surface resolver (`build_vocabulary_surface_candidates` in `scripts/audit/post_build_review.py`) split stressed learner tokens mid-word — U+0301 is not a `\w` word char, so `м'яки́й` tokenized as `м'яки` + `й`, the multi-token whitespace-gap check then rejected every stressed window — and compared raw bytes for exact/VESUM matching. Stressed a1/special-signs surfaces therefore missed otherwise valid forms and produced false MISSING findings in canonical post-build review.

Fix, scoped to candidate matching/VESUM lookup only:

- `_LEXICAL_TOKEN_PATTERN` admits U+0301 inside word tokens (`scripts/audit/post_build_review.py:2012-2014`).
- New comparison-only `_candidate_match_key` strips U+0301 before exact matching, VESUM lookup keys, and VESUM lemma comparison (`:2021-2023`, call sites `:2147-2155`); `_contains_exact_token_sequence` compares via the same key so packet-bound stressed surfaces pass coverage validation.
- `resolver_version` bumped `vesum-surface-candidates.v1` → `v2` (informational; nothing validates it).

**Displayed/stored evidence bytes are unchanged**: candidate `surface` slices and `VESUM:` verification strings keep the original stressed bytes (asserted in tests).

## Verification (all commands run in the dispatch worktree)

1. False-MISSING repro → match. Real a1/special-signs ledger + module + activities (from `origin/main` object store) + real VESUM DB, before/after quoted:

   ```
   # before (main): MISSING (21/29): "м'яки́й знак", 'апо́строф', "м'яки́й", 'тверди́й', ...
   # after (this PR): MISSING (3/29): 'зли́то', "здоро́в'я", 'сього́дні'
   ```
   The remaining three stems never occur in the module's learner-facing text (verified stress-insensitively over module.md + activities.yaml) — a genuine content gap, out of scope per the issue's no-content-edits constraint.

2. Regression tests added (`tests/audit/test_post_build_review.py`):
   - `test_vocabulary_candidates_match_across_combining_stress_marks` — stressed↔unstressed exact match, stressed inflected VESUM match with raw bytes preserved, lexical drift (`злиття` vs `зли́то`) still rejected.
   - `test_vocabulary_coverage_accepts_stressed_exact_surface` — packet-bound stressed surface passes coverage validation.

3. Mutation check (#M-16): with the fix stashed (`git stash push -- scripts/audit/post_build_review.py`):
   ```
   FAILED tests/audit/test_post_build_review.py::test_vocabulary_candidates_match_across_combining_stress_marks
   FAILED tests/audit/test_post_build_review.py::test_vocabulary_coverage_accepts_stressed_exact_surface
   2 failed, 202 deselected
   ```
   Both fail without the fix, pass with it.

4. `.venv/bin/pytest tests/audit/test_post_build_review.py`: `53 failed, 105 passed, 48 errors` — the failing set is byte-identical (`diff` verified) to clean `origin/main` in this sparse checkout (`curriculum/` absent; pre-existing environmental failures, not regressions). `.venv/bin/ruff check` on both changed files: all checks passed.

Sibling PR for #5374 (apostrophe glyphs) stacks on this branch.

### DIFF
--- a/scripts/audit/post_build_review.py
+++ b/scripts/audit/post_build_review.py
@@ -2003,14 +2003,34 @@ def _visible_source_lines(path: str, text: str) -> list[str]:
     return masked.splitlines()
 
 
+# Combining acute stress marks (U+0301) are pedagogical annotations, not
+# lexical identity: a stressed learner surface (м'яки́й) and its unstressed
+# ledger lemma (м'який) are the same word. Word tokens must admit the mark
+# mid-token (м'яки́й otherwise splits into м'яки + й and can never match), and
+# comparisons strip it — for candidate matching and VESUM lookup only. Stored
+# and displayed evidence always keeps the original bytes. (#5375)
+_COMBINING_STRESS_MARK = "\u0301"
+_TOKEN_WORD = "[^\\W_]+(?:\u0301+[^\\W_]*)*"
+_LEXICAL_TOKEN_PATTERN = re.compile(rf"{_TOKEN_WORD}(?:[’']{_TOKEN_WORD})*", flags=re.UNICODE)
+
+
+def _strip_combining_stress(text: str) -> str:
+    return text.replace(_COMBINING_STRESS_MARK, "")
+
+
+def _candidate_match_key(text: str) -> str:
+    """Comparison-only key for candidate resolution; never applied to evidence bytes."""
+    return _strip_combining_stress(text).casefold()
+
+
 def _lexical_tokens(value: str) -> list[str]:
     """Return Unicode word tokens without inferring any morphology."""
-    return [token.casefold() for token in re.findall(r"[^\W_]+(?:[’'][^\W_]+)*", value, flags=re.UNICODE)]
+    return [token.casefold() for token in _LEXICAL_TOKEN_PATTERN.findall(value)]
 
 
 def _contains_exact_token_sequence(value: str, phrase: str) -> bool:
-    value_tokens = _lexical_tokens(value)
-    phrase_tokens = _lexical_tokens(phrase)
+    value_tokens = [_candidate_match_key(token) for token in _lexical_tokens(value)]
+    phrase_tokens = [_candidate_match_key(token) for token in _lexical_tokens(phrase)]
     if not phrase_tokens:
         return False
     width = len(phrase_tokens)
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Cross-family review of record — VERDICT: REQUEST_CHANGES

Reviewer: Claude Fable (inline; cross-family vs author kimi/Moonshot; language-lane seat for the orthography facts), head 86fee6b5cb.

**Verified correct:** U+0301 as pedagogical stress annotation stripped for matching only (м'яки́й ≡ м'який) with evidence bytes preserved — linguistically right and properly tested; the three-glyph apostrophe equivalence set (U+0027/U+2019/U+02BC, U+02BB stays distinct) matches the #5374 contract; token pattern correctly admits U+0301 mid-token so stressed words no longer split.

**Blocker — U+02BC never reaches the fold on tokenizer paths:** `_APOSTROPHE_FOLD` is applied per-token AFTER tokenization, but `_LEXICAL_TOKEN_PATTERN`'s joiner class is `[’']` — U+02BC (ʼ) is not in it. A source surface written with the modifier-letter apostrophe (`мʼякий` — the Unicode-recommended form some corpora/keyboards produce, and named explicitly in #5374) tokenizes as `м` + `який`, so the fold never sees the joined word and the false-MISSING this issue exists to kill still occurs on every tokenizer call site. Fix: add ʼ to the joiner class (`[’'ʼ]`) + a regression test with a U+02BC-carrying surface through the tokenizer path (not only the whole-string `_candidate_match_key` path), mutation-checked.

Fix returns to the authoring lane on this branch. Everything else is merge-ready.

krisztiankoos:
Correction (orchestrator): the review above was run against the combined two-commit diff; this PR is the #5375 stress-mark scope ONLY (head 203fffeb7) — the apostrophe work and the U+02BC blocker belong to PR #6833. For THIS PR's scope the verdict is **APPROVE**: U+0301 stripped for matching only, evidence bytes preserved, token pattern admits the mark mid-token, tests pin it. The U+02BC joiner gap is not a regression of this diff (the joiner class predates it) and is tracked as #6833's blocker. Arming auto-merge.

PR #6833, opened 2026-08-15
fix(review): fold Ukrainian apostrophe glyphs in vocabulary surface candidates (#5374)

krisztiankoos:
Closes #5374.

**Stacked on #6832** (the #5375 stress-mark PR): the two fixes share the same comparison-only match key in the same resolver function, so this branch contains that commit; the diff shown here is only the apostrophe fold. GitHub will retarget this PR to `main` once #6832 merges. Per the dispatch brief, one PR per issue was preferred over a combined PR.

## What

The vocabulary surface resolver treated the three Ukrainian apostrophe glyphs — U+0027 (`'`), U+2019 (`’`), U+02BC (`ʼ`) — as distinct for exact matching and VESUM lookup, even though they are the same orthographic sign (the semantic claim binder already folds them) and VESUM stores the ASCII form. A glyph-mixed learner text produced false MISSING vocabulary findings.

Fix: `_APOSTROPHE_FOLD` folds U+2019 and U+02BC to U+0027 inside the comparison-only `_candidate_match_key` (`scripts/audit/post_build_review.py`), which feeds exact matching, VESUM lookup keys, VESUM lemma comparison, and `_contains_exact_token_sequence` coverage validation. No tokenizer change was needed — U+02BC is a letter (Lm) and already tokenizes inside words.

- **Evidence bytes preserved**: candidate surfaces and `VESUM:` verification strings keep the original glyph (asserted in tests).
- **Glyphs outside the three stay distinct**: U+02BB (`ʻ`) is never folded and produces no candidate (asserted in tests).

## Verification (dispatch worktree)

1. False-MISSING repro → match. Real a1/special-signs ledger (ASCII apostrophes) against a glyph-mixed copy of the module content (apostrophes alternately swapped to U+2019/U+02BC), real VESUM DB:

   ```
   pre-fix MISSING on glyph-mixed content: 14/29
     - "м'яки́й знак", "сім'я́", "м'я́со", "п'ять", "де́в'ять", "ім'я́", "комп'ю́тер", "бур'я́н", "Мар'я́на", "дерев'я́ний", ...
   post-fix MISSING on glyph-mixed content: 3/29
     - 'зли́то', "здоро́в'я", 'сього́дні'   (genuinely absent from learner text — same residual as unmixed content)
   ```

2. Regression tests added (`tests/audit/test_post_build_review.py`):
   - `test_vocabulary_candidates_fold_ukrainian_apostrophe_glyphs` — U+02BC and U+2019 surfaces exact-match the U+0027 lemma; U+2019 inflected surface matches via VESUM; asserts the VESUM lookup received the folded ASCII form; asserts U+02BB binds nothing; surfaces keep original bytes.
   - `test_vocabulary_coverage_accepts_modifier_apostrophe_exact_surface` — packet-bound U+02BC surface passes coverage validation for the U+0027 lemma.

3. Mutation check (#M-16): with `.translate(_APOSTROPHE_FOLD)` removed from the match key:
   ```
   FAILED tests/audit/test_post_build_review.py::test_vocabulary_candidates_fold_ukrainian_apostrophe_glyphs
   FAILED tests/audit/test_post_build_review.py::test_vocabulary_coverage_accepts_modifier_apostrophe_exact_surface
   2 failed, 9 passed
   ```
   Both fail without the fold, pass with it; the #5375 stress tests stayed green throughout (no over-broadening).

4. `.venv/bin/pytest tests/audit/test_post_build_review.py`: failing set byte-identical (`diff` verified) to clean `origin/main` in this sparse checkout (pre-existing `curriculum/`-absence failures, not regressions); the four new tests pass. `.venv/bin/ruff check` on both changed files: all checks passed.

No `scripts/lexicon/` files touched — the Atlas manifest freshness gate does not apply.

### DIFF
--- a/scripts/audit/post_build_review.py
+++ b/scripts/audit/post_build_review.py
@@ -2011,7 +2011,10 @@ def _visible_source_lines(path: str, text: str) -> list[str]:
 # and displayed evidence always keeps the original bytes. (#5375)
 _COMBINING_STRESS_MARK = "\u0301"
 _TOKEN_WORD = "[^\\W_]+(?:\u0301+[^\\W_]*)*"
-_LEXICAL_TOKEN_PATTERN = re.compile(rf"{_TOKEN_WORD}(?:[’']{_TOKEN_WORD})*", flags=re.UNICODE)
+# U+02BC is a modifier letter (Lm), so _TOKEN_WORD already matches it
+# mid-word; it is named in the joiner class explicitly so the apostrophe
+# equivalence set stays visibly complete with _APOSTROPHE_FOLD below. (#5374)
+_LEXICAL_TOKEN_PATTERN = re.compile(rf"{_TOKEN_WORD}(?:[’'ʼ]{_TOKEN_WORD})*", flags=re.UNICODE)
 
 # The three Ukrainian apostrophe glyphs — U+0027 ASCII apostrophe, U+2019
 # right single quotation mark, and U+02BC modifier letter apostrophe — are the
--- a/scripts/audit/post_build_review.py
+++ b/scripts/audit/post_build_review.py
@@ -2013,6 +2013,13 @@ _COMBINING_STRESS_MARK = "\u0301"
 _TOKEN_WORD = "[^\\W_]+(?:\u0301+[^\\W_]*)*"
 _LEXICAL_TOKEN_PATTERN = re.compile(rf"{_TOKEN_WORD}(?:[’']{_TOKEN_WORD})*", flags=re.UNICODE)
 
+# The three Ukrainian apostrophe glyphs — U+0027 ASCII apostrophe, U+2019
+# right single quotation mark, and U+02BC modifier letter apostrophe — are the
+# same orthographic sign; VESUM stores the ASCII form. Fold them for
+# candidate comparison and VESUM lookup only, never in stored or displayed
+# evidence. Glyphs outside these three (e.g. U+02BB) stay distinct. (#5374)
+_APOSTROPHE_FOLD = str.maketrans({"’": "'", "ʼ": "'"})
+
 
 def _strip_combining_stress(text: str) -> str:
     return text.replace(_COMBINING_STRESS_MARK, "")
@@ -2020,7 +2027,7 @@ def _strip_combining_stress(text: str) -> str:
 
 def _candidate_match_key(text: str) -> str:
     """Comparison-only key for candidate resolution; never applied to evidence bytes."""
-    return _strip_combining_stress(text).casefold()
+    return _strip_combining_stress(text).translate(_APOSTROPHE_FOLD).casefold()
 
 
 def _lexical_tokens(value: str) -> list[str]:
--- a/tests/audit/test_post_build_review.py
+++ b/tests/audit/test_post_build_review.py
@@ -3258,6 +3258,47 @@ def test_vocabulary_candidates_fold_ukrainian_apostrophe_glyphs() -> None:
     assert all("\u02bb" not in candidate["surface"] for candidate in by_lemma["сім'я"])
 
 
+def test_vocabulary_candidates_resolve_modifier_apostrophe_stressed_token() -> None:
+    """#5374 review fix: a U+02BC surface carrying the combining stress mark
+    (Мʼяки́й знак) must stay ONE token on the tokenizer path — if the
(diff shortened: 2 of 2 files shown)


krisztiankoos:
## Cross-family review of record — VERDICT: REQUEST_CHANGES

Reviewer: Claude Fable (inline; cross-family vs author kimi/Moonshot; language-lane seat), reviewing the apostrophe-fold scope (#5374) at head 86fee6b5cb.

**Verified correct:** three-glyph equivalence set (U+0027/U+2019/U+02BC; U+02BB stays distinct) matches the issue contract; fold applied for comparison/VESUM lookup only; evidence bytes preserved.

**Blocker (full detail in the review comment on #6832, posted before the two-PR split was visible):** `_APOSTROPHE_FOLD` runs per-token AFTER tokenization, but the tokenizer joiner class is `[’']` — U+02BC (ʼ) absent. A U+02BC surface (`мʼякий`) splits into `м`+`який` before the fold applies, so the #5374 false-MISSING persists on every tokenizer path. Fix: add ʼ to the joiner class + a mutation-checked regression with a U+02BC surface through the tokenizer path. The authoring lane is already dispatched on this branch (task infra-5374-joiner-fix). Merge holds for the fix + re-check; recommend updating this branch after #6832 merges.

krisztiankoos:
**Blocker addressed at head 869c1ff92b** — with one evidence-based correction on the mechanism.

**Fix applied:** the joiner class is now `[’'ʼ]` in `_LEXICAL_TOKEN_PATTERN` (scripts/audit/post_build_review.py:2017), with a comment recording why. New regression test `test_vocabulary_candidates_resolve_modifier_apostrophe_stressed_token` drives `Мʼяки́й знак` (U+02BC + U+0301) through the tokenizer path (`build_vocabulary_surface_candidates`) and asserts it resolves as `("Мʼяки́й знак", "exact lemma surface")` against ledger lemma `м'який знак`, evidence bytes preserved.

**Correction on the blocker mechanism (tool-backed):** the false-MISSING did not actually persist on tokenizer paths. U+02BC is Unicode category **Lm** (modifier letter), so `[^\W_]` in `_TOKEN_WORD` already matches it mid-word — the token never split. Verified at the reviewed head 86fee6b5cb:

```
>>> unicodedata.category("ʼ")
'Lm'
>>> pbr._lexical_tokens("Мʼяки́й знак")
['мʼяки́й', 'знак']
>>> pbr._contains_exact_token_sequence("Тут Мʼяки́й знак стоїть.", "м'який знак")
True
```

The existing `test_vocabulary_candidates_fold_ukrainian_apostrophe_glyphs` already proved this end-to-end: its `Сімʼя` surface reaches `"exact lemma surface"` for lemma `сім'я` only via the tokenizer path at line 2125 — it passed at 86fee6b5cb.

**Mutation check (#M-16), quoted honestly:**
- Revert the joiner change (`[’'ʼ]` → `[’']`), run both apostrophe tests: `2 passed` — the joiner entry is explicitness, not the load-bearing guard; the requested mutation does not bite.
- The load-bearing mutation — exclude U+02BC from `_TOKEN_WORD` (`[^\W_]` → `[^\W_ʼ]`), joiner still reverted: both tests **FAILED** with `м` + `яки́й` token split, exactly the false-MISSING described. The new test pins the real guarantee.

**Checks:** ruff check + format green on both changed files. `pytest tests/audit/test_post_build_review.py -q` in the dispatch worktree: 106 passed, 53 failed, 48 errors — the failure set is byte-identical at HEAD (105 passed baseline; all 101 are sparse-checkout artifacts: the worktree lacks the `curriculum/` and `wiki/` trees, e.g. `FileNotFoundError: curriculum/l2-uk-en/plans/bio/oleksandr-bilash.yaml`). Zero regressions introduced; full-checkout CI is the green denominator.

krisztiankoos:
Correction (orchestrator, tool-verified): my blocker's failure scenario was WRONG. U+02BC is Unicode category Lm — a word character under `[^\W_]` — so `мʼякий` tokenized as ONE token under the pre-existing pattern and the per-token fold already normalized it (verified: `old-pattern tokens: ['Мʼяки́й', 'знак']`). The authoring lane's note caught this. The shipped change (explicit ʼ in the joiner class + U+02BC regression test) is correct hardening that keeps the equivalence set visibly complete, and costs nothing.

**VERDICT: APPROVE** at head 869c1ff92. Arming auto-merge.

Issue #6515, opened 2026-08-09
[harness][sources] Expose stress position (stressed form + vowel index) through the sources MCP — blocks #4696 TTS bake-off


krisztiankoos:
## Why
The #4696 voice research (report on that issue, 2026-08-09) found the live sources/VESUM tools accept unaccented forms only, reject U+0301 stress-marked variants, and return lemma/POS/tags without stress position. The TTS stress-correctness protocol needs a stress oracle: for a given lemma/form, the stressed form and stressed-vowel index, bound to the VESUM identity.

## Ask (harness lane)
- Expose stressed form + vowel index via the sources MCP (project stress dictionary already used by the stress deck).
- Accept NFC U+0301 input without erroring (normalize or verify).
- Versioned output (source digest) so bake-off records are reproducible.

## Consumer
#4696 200-clip TTS bake-off (operator-gated) — scoring is impossible without this.

Refs #4387, #4913.

X-Agent: claude-atlas

krisztiankoos:
## Design note — stress oracle for the sources MCP

Investigated the live stress code paths and data assets before proposing anything (evidence below is tool-verified against this worktree, not recalled). Design only — no code in this comment.

### 1. Data inventory

**Primary local source — `ukrainian_word_stress` (pip package), already a project dependency**
- Backing data: `ukrainian_word_stress/data/stress.trie`, a `marisa_trie` compiled from ULIF's external `ukrainian-word-stress-dictionary` project (`ulif_accents.csv`), per `compile_dict.py`'s `__main__`. This is a *different* pipeline from the live `mcp__sources__query_ulif` tool below — same institution, unrelated code path.
- Verified size: **2,892,732 word-form entries**; **37,212 (1.29%) carry more than one valid stress** and need POS/case/number tags to disambiguate (heteronyms, e.g. за́мок/замо́к) — measured directly against the installed trie, not the README.
- Disambiguation: `Disambiguation.Stanza` (default when Stanza is installed — POS-context aware, ~500MB model) vs `Disambiguation.Dictionary` (`on_ambiguity`: `skip`/`first`/`all`).
- **Version drift, confirmed**: `requirements.txt` says `>=1.0`, `requirements-lock.txt` pins `==1.1.1`, but the active `.venv` resolved **`2.1.0`**. The trie shape (and thus every stress answer) changes across releases — an oracle built today would silently answer from an unpinned version.
[...]

Issue #4696, opened 2026-07-07
[4387][practice][USER-GATED] Pronunciation audio — TTS vs recorded voice decision + pipeline


krisztiankoos:
Biggest missing learner value: zero audio anywhere (flashcards/atlas have stress+IPA but no sound; listening mode disabled by design pending assets).

USER DECISION NEEDED before any build: (a) TTS (which engine — quality/licensing for a non-commercial open resource; ukrainian-tts/espnet options), vs (b) recorded native audio (cost/effort/volunteers), vs (c) defer. Scope if approved: build-time audio asset pipeline (versioned, like deck assets), player on flashcards + atlas headwords, later listening mode (spec has it designed-but-disabled).

Refs #4387. DO NOT start without explicit user go.

krisztiankoos:
## Disposition — named hold `human_decision`

Pronunciation audio remains user-gated. No dispatch.

X-Agent: grok-atlas

krisztiankoos:
Operator dispatch on 2026-09-05 explicitly un-gated #4696 and ordered local TTS, a bounded first slice, flashcard + Atlas players, one PR and no merge. This supersedes the earlier hold for this dispatched scope.

Engine installed and exercised: Piper 1.3.0 (GPL-3.0-or-later), `uk_UA-ukrainian_tts-medium`, speaker `lada`/0; model repository MIT and voice dataset CC0. Model revision and both file hashes are pinned. A first real local run generated 200 WAVs from the hydrated A1 deck's 1,371 source rows. Bulk WAVs and engine/model caches are ignored; a tiny WAV fixture is tracked.

Cross-family design review: Claude Sonnet hit its session limit; substituted Grok 4.6. Material recommendations being incorporated: align Atlas lookup normalization with the deck; include monosyllables without inventing a stress position; preserve hydrated deck order; put the Atlas control outside the heading. Playback tests, SSR lifecycle handling, Pages generation and runbook are now implemented. Exact-head implementation review follows the PR.

Structural audio validation is not pronunciation certification. Listening remains disabled; held-out Ukrainian listening/stress evaluation remains a separate residual owned by the #4696 integration/operator lane. This PR will not close that evidence gap or claim whole-Atlas coverage.

X-Agent: codex/4696-tts-pipeline

krisztiankoos:
PR #7717 verification update at `fdfee6268243e2eec2e3ea64ff57a926e6ee1372`:

[...]

PR #6836, opened 2026-08-15
feat(sources): add verify_stress oracle (#6515)

krisztiankoos:
## What

New `mcp__sources__verify_stress` tool: stressed form + stressed-vowel index for a Ukrainian word, implementing the design pinned on #6515 (design note + K3 cross-family review + round-2 revision — the round-2 revision is the spec this PR implements).

## Evidence (#M-4)

- `scripts/verification/stress.py` — the oracle. Trie-decode logic (`_parse_dictionary_value`/`_apply_accents`) validated against the package's own `Stressifier` output across ~3000 sampled trie entries before writing tests (0 mismatches) — see PR conversation for the verification script.
- Golden-entry test `tests/test_stress_oracle.py::TestStressGolden::test_zamok_golden_entry` pins the exact 5-reading decode of `замок`'s packed trie value (round-2 §3c safety net).
- Live one-shot smoke through the actual MCP dispatch path (`call_tool("verify_stress", {"word": "замок"})`) — full JSON envelope confirmed, 5 enumerated readings incl. `unresolvable_by_tags: true` for the true castle/lock homograph.
- `.venv/bin/pytest tests/test_stress_oracle.py tests/test_mcp_sources_server.py -q` → **74 passed**, 3 pre-existing failures (`ModuleNotFoundError: rapidfuzz`, confirmed unrelated — reproduces identically on `git stash` of this PR's diff), 5 skipped (pre-existing VESUM-DB-dependent smoke tests).
- `.venv/bin/ruff check` clean on all changed Python files.
- Mutation-checked (#M-16): reverting the override-layer line breaks 2/3 `TestOverrides` tests; reverting the `input_mismatch` comparison breaks `test_disagreeing_mark_is_true`. Both pass again after restore, diff confirmed identical to pre-mutation.

## Contract implemented (round-2 revision, verbatim points)

- §2: `ukrainian-word-stress` pinned `requirements.txt` `>=2,<3`, `requirements-lock.txt` `==2.1.0` (was drifted `1.1.1`/`>=1.0`) + `source.digest` = sha256 of the loaded trie file.
- §3a: `input_mismatch` per-match, omitted (not `false`) when the caller's input carried no stress mark.
- §3b: `required_tags` uses the dictionary's own native `upos=*` vocabulary; `unresolvable_by_tags` flags true meaning-dependent homographs (byte-identical `required_tags` across distinct stress positions) that no tag combination can disambiguate.
- §3c: packed-value byte-layout parsing sanctioned, gated by version pin + trie digest + the golden-entry test.
- §3d: new `invalid_input` status (empty/multi-word/non-Cyrillic/single-syllable), distinct from `not_found`.
- §4: override layer applied by exact-lemma-string match only (never inflected forms) — matches `generate_ipa.py`'s existing behavior; VESUM join uses `scripts/verification/vesum.py::verify_word` (worktree-fallback already solved via `scripts/rag/config.py`, #6542).
- Registration: `Tool` + dispatch-dict + `handle_verify_stress` in `.mcp/servers/sources/server.py`, same three touch points as `verify_word`/`verify_lemma`.

## Extension beyond the pinned schema (found during implementation, documented in the module docstring)

Hyphenated compound entries (e.g. `попереково-крижовими`) can carry more than one stress mark in a single trie value (~6.9K/2.9M entries, ~0.24%) — not addressed by either design round. `vowel_index` stays a single int for schema compatibility (first/lowest mark); `vowel_indices` (always present) carries the full list so compound stress isn't silently dropped.

## Out of scope (explicit)

`query_ulif` and mphdict live/hydration-dependent cross-checks — the design note frames both as optional "keep the door open" extensions, not pinned deliverables. The trie stays the sole source of truth; nothing in this PR depends on either.

Refs #6515.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/.mcp/servers/sources/server.py
+++ b/.mcp/servers/sources/server.py
@@ -16,6 +16,7 @@ with imports; rename there is a follow-up.
 Tools:
     - search_sources, search_text, search_literary, search_external, search_images, get_chunk_context
     - verify_word, verify_words, verify_lemma, vet_vocabulary (VESUM)
+    - verify_stress (stress oracle: ukrainian-word-stress trie + override layer + VESUM join)
     - query_wikipedia, query_pravopys, query_e2u, query_r2u, query_ulif
     - search_definitions, search_grinchenko_1907, search_esum, search_idioms, search_synonyms
     - search_slovnyk_me, search_heritage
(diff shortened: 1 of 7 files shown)


krisztiankoos:

Reviewer: Kimi K3 (task review-6836; the constraint author for this design), head 24ee01316e, reviewed against the pinned round-2 spec in a scratch worktree at the exact head. Highlights: 23/23 oracle + 54 server tests run by the reviewer itself; both quoted mutation-checks REPRODUCED (not just assessed); independent full-scan of all 2,892,732 trie entries proving the value-parser fast path cannot misread tag bytes; contract fidelity verified per-point with file:line (input_mismatch omitted-not-false; dictionary-native upos=* vocabulary per the spec's evidence-backed position; sanctioned enumeration via public symbols + version/digest safety net; error-class separation; exact-lemma override semantics; VESUM one-layer-down fallback; version pins byte-identical to spec; registration at all three server touch points).

Non-blocking follow-up candidates recorded in the review transcript: partial-tag narrowing semantics for required_tags, trailing-space leniency, one cosmetic test placement. Pre-existing ruff I001 in server.py reproduces on main — not this PR.

This closes the #6515 chain: design → CF review → evidence-based revision → implementation → CF implementation review. The #4696 TTS bake-off consumer remains operator-gated. Merging on green CI.

Issue #6823, opened 2026-08-14
[atlas][a11y] Word-page hero lemma is near-black on the teal band


krisztiankoos:
## What
Atlas word pages paint the lemma as near-black on the dark teal hero. The gloss line under it is correctly white. Contrast is about 1.5:1. Every Atlas entry is affected.

Live 2026-08-15 after Pages #1062 (PR #6819) on https://learn-ukrainian.github.io/lexicon/%D0%BF%D0%BE%D0%B4%D0%B0%D0%B2%D0%B0%D1%82%D0%B8%D1%81%D1%8F/

Computed (light theme):
- `h1.word-title` «подаватися» → `rgb(33,33,33)` (`--lu-text`) over `linear-gradient(135deg,#00695C,#004D40)`
- sibling gloss `дієслово · «to be submitted; to be served»` → `#FFFFFF` (readable)
- `.word-stress` `[подавáтися]` inherits the dark title color

## Why (specificity, not missing CSS)
Live `CourseLayout` CSS already has `.word-title{color:#fff}`, but a later/stronger rule wins:

```css
.lu-content h1, .lu-content h2, .lu-content h3 { color: var(--lu-text); }
```

`.lu-content h1` (0,1,1) beats `.word-title` (0,1,0).

Same pattern on the hero CTA: `.lu-content .word-atlas a { color: var(--blue); }` (0,2,1) beats `.word-hero .practice-cta-hero { color:#fff }` (0,2,0), so «Practice this word →» is `#0057B7` on teal (also low contrast, slightly less severe).

This is **not a #6819 regression**. The same dark-on-teal title was already live after Pages #1060. The #5376 axe gate only scans `/lexicon/` landing + practice, so word pages never tripped it. Filing now because the contrast work in #6814/#6815 made this leftover obvious.

## Expected
Title, stress form, and hero CTA stay white (or otherwise WCAG AA) on the teal band. A selector that actually wins, e.g. `.word-hero .word-title` / `.word-hero .practice-cta-hero`, not a new token rename.

## Repro
1. Open https://learn-ukrainian.github.io/lexicon/%D0%BF%D0%BE%D0%B4%D0%B0%D0%B2%D0%B0%D1%82%D0%B8%D1%81%D1%8F/ (light theme)
2. Look at the teal hero: lemma is dark, gloss is white

## Severity
Medium. The word the page is about is the least readable thing on it.

Related: #6814 #6815 #5376

[...]

PR #6838, opened 2026-08-15
fix(atlas): word-page hero lemma/CTA contrast on teal (#6823)

krisztiankoos:
## Summary
- Raise specificity so the teal word-hero keeps a white lemma, stress form, and Practice CTA under CourseLayout content chrome (`#6823`).
- Add a happy-dom regression test that replays the live cascade fight and quotes computed colors (`#fff` after; legacy selectors lose to `--lu-text` / `#0057B7`).

### Selector before → after
| Surface | Before (lost) | After (wins) |
| --- | --- | --- |
| Title | `.word-title` (0,1,0) vs `.lu-content h1` (0,1,1) | `.word-hero .word-title` (0,2,0) |
| Stress | inherited losing title color | `.word-hero .word-stress` with `color: #fff` |
| CTA | `.word-hero .practice-cta-hero` (0,2,0) vs `.lu-content .word-atlas a` (0,2,1) | `.lu-content .word-atlas .word-hero .practice-cta-hero` (0,4,0) |

### Computed colors (happy-dom, production cascade)
- `h1.word-title` → `#fff` (was live `rgb(33,33,33)` / `--lu-text`)
- `.word-stress` → `#fff`
- `a.practice-cta-hero` → `#fff` (was live primary blue on teal)

Light + dark themes covered (teal band still present under `[data-theme=dark]`).

## Test plan
- [x] `cd site && npx vitest run tests/unit/atlas-hero-contrast.test.ts` (3/3 pass)
- [ ] Spot-check `/lexicon/подаватися/` hero title + CTA after Pages deploy
- [ ] Do not auto-merge — leave `#6823` open until CF + merge

Closes nothing (issue stays open until CF + merge per dispatch).

### DIFF
--- a/site/src/styles/word-atlas.css
+++ b/site/src/styles/word-atlas.css
@@ -330,19 +330,22 @@
   text-transform: uppercase;
 }
 
-.word-title {
+/* Must beat `.lu-content h1` (0,1,1) from course.css — bare `.word-title`
+   (0,1,0) loses and paints --lu-text on the teal hero (#6823). */
+.word-hero .word-title {
   margin: 0 0 4px;
-  color: white;
+  color: #fff;
   font-size: 44px;
   font-weight: 700;
   line-height: 1.1;
 }
 
-.word-stress {
+.word-hero .word-stress {
   margin-left: 14px;
   font-family: "SF Mono", Menlo, monospace;
   font-size: 26px;
   font-weight: 400;
+  color: #fff;
   opacity: 0.85;
 }
 
@@ -360,8 +363,10 @@
   align-items: center;
 }
 
-/* Atlas → Practice CTA sits in the teal word hero; keep white chrome for contrast. */
-.word-hero .practice-cta-hero {
+/* Atlas → Practice CTA sits in the teal word hero; keep white chrome for contrast.
+   Must beat `.lu-content .word-atlas a` (0,2,1) — bare `.word-hero .practice-cta-hero`
+   (0,2,0) loses and paints --blue on teal (#6823). */
+.lu-content .word-atlas .word-hero .practice-cta-hero {
   display: inline-flex;
   align-items: center;
   min-height: 2rem;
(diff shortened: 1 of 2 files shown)


krisztiankoos:
VERDICT: APPROVE

Cross-family review (kimi/review-6838) at head `1ed3d92`. Read both changed files in full, verified the production markup chain, ran the test at the PR head, and ran it against `origin/main`'s old CSS as a mutation check.

## 1. Selector specificity — quoted and verified

- **Title:** `.word-hero .word-title` = **(0,2,0)** vs `.lu-content h1` (`course.css:559`) = **(0,1,1)**. Class count compared first: 2 > 1 → **yes, it beats it**, independent of source order.
- **CTA:** `.lu-content .word-atlas .word-hero .practice-cta-hero` = **(0,4,0)** vs `.lu-content .word-atlas a` (`word-atlas.css:98`) = **(0,2,1)**. 4 > 2 → **yes, it beats it**.
- **Stress:** `.word-hero .word-stress` (0,2,0) now carries explicit `color: #fff`; the span is not an h-element so `.lu-content h1` never matched it directly — the fix correctly targets the inherited losing color.
- Hover (`(0,5,0)` vs `.word-atlas a:hover` (0,2,1)), `--disabled`, dark-theme, and the mobile media-query variants were all re-scoped consistently — no orphaned bare selectors remain (grep confirms `.word-title`/`.word-stress`/`practice-cta-hero` appear only under the new scoped forms).

Production nesting confirmed: `CourseLayout.astro:381` renders `.lu-content`, `WordAtlasArticle.tsx:146` renders the `.word-atlas` root with `.word-hero` wrapping `h1.word-title`, `.word-stress`, and `a.practice-cta-hero` — so the (0,4,0) CTA selector really matches in production, and `[lemma].astro` uses `CourseLayout`.

## 2. Test fails on the old rule order — confirmed empirically

- At PR head: `npx vitest run tests/unit/atlas-hero-contrast.test.ts` → **3/3 pass**.
- With `site/src/styles/word-atlas.css` swapped to `origin/main`'s version: **2 failed, 1 passed** — the source-contract test fails its regex assertion and the computed-color test errors with `Missing CSS rule for selector: .word-hero .word-title`. (File restored afterward; worktree is clean at `1ed3d92`.)
- The third test is a built-in mutation guard that replays the pre-fix cascade (bare `.word-title` + `.word-hero .practice-cta-hero` against `.lu-content h1` / `.lu-content .word-atlas a`) and asserts the legacy selectors lose — so regression protection is in the test itself, not just in my swap.

## Findings (non-blocking)

- Minor: on the old CSS the computed-color test fails via a "Missing CSS rule" error rather than a color-value assertion. Still a hard failure, so protection holds — just slightly less direct signal. Not worth a respin.
- The test reconstructs a cascade slice by extracting individual production rules rather than loading full stylesheets. The extracted set covers every competitor named in the bug (`:root`/`dark` tokens, `.lu-content h1`, `.lu-content a`, `.lu-content .word-atlas a`), and specificity — not order — decides all three fights, so the slice is faithful for this regression.

Fix is minimal, comments document exactly why the specificity is needed, and the stated live-bug mechanism matches what I verified in the sources. APPROVE — leaving merge to the author lane per dispatch.

Issue #6812, opened 2026-08-14
[infra][tests] Bridge loads under two module identities (ai_agent_bridge vs scripts.ai_agent_bridge) — standardize on one


krisztiankoos:
Surfaced by the #6808 review (Kimi seat, tool-backed): `scripts.ai_agent_bridge` is imported in ~199 places while 16+ test files import top-level `ai_agent_bridge` — including three files inside `tests/ai_agent_bridge/` that use the scripts-prefixed form while siblings use the bare form. Both identities load as separate `sys.modules` entries with separate module-level state (e.g. the `_db` session/engine helpers at `scripts/ai_agent_bridge/_db.py:509`). No single file mixes both styles today, so tests stay self-consistent — but `monkeypatch` on one identity never affects the other, and cross-file state assumptions can silently diverge.

Recommendation from the review: standardize the whole package on `scripts.ai_agent_bridge` (repo root is already on sys.path via `tests/conftest.py`), which would eventually allow deleting ~340 inline `sys.path.insert` primers across the test suite. Pre-existing debt — not introduced by #6808, which only made the existing behavior deterministic.

X-Agent: claude-atlas

krisztiankoos:
Reopened: the first repair standardized test imports but API startup preload still imports both package identities.

Tool-backed reproduction at `c66fa8d8b22638d003bc2990cf80e149b3adb1a8`:

```text
pytest tests/api/test_import_pinning.py::test_lifespan_preload \
  tests/ai_agent_bridge/test_module_identity.py::test_bare_bridge_identity_not_loaded -q
1 passed, 1 failed
```

The failure is the existing `test_bare_bridge_identity_not_loaded` assertion. `scripts/api/preload.py::OPTIONAL_MODULES` explicitly lists the three bare bridge module names before the canonical `scripts.ai_agent_bridge` names. This is also the sole pytest failure on PR #6842 (3,967 passed) and on main run 31908853114.

A narrow follow-up will canonicalize only the preload entries and add an import-pinning regression test. It will not weaken the identity guard or change bridge runtime behavior.

X-Agent: codex/work-control-plane

krisztiankoos:
Shipped in #6844 (`e2d3aeeaad`). Monitor API preload and comms sweep use `scripts.ai_agent_bridge` only. Pytest shard 1 passed on that PR.

X-Agent: grok-atlas

PR #6844, opened 2026-08-16
fix(api): load bridge via scripts.ai_agent_bridge only (#6812)

krisztiankoos:
## Summary
Root cause of the xdist flake that blocked #6838/#6840/#6841/#6843: `scripts/api/preload.py` and `comms_router.py` imported the bare `ai_agent_bridge` identity. The Monitor API boots inside pytest, so those imports created a second `sys.modules` entry depending on shard order.

Scan `scripts/api/` in `test_module_identity.py`. Do not weaken the single-identity contract.

Finalized in the existing `kimi/atlas-6812-identity-flake` worktree after the worker hit hard_timeout mid full-suite. Targeted tests: 46 passed. No new worktree.

X-Agent: grok-atlas

### DIFF
--- a/scripts/api/comms_router.py
+++ b/scripts/api/comms_router.py
@@ -200,10 +200,15 @@ def _maybe_run_delivery_expiry_sweep() -> None:
 
 def _run_delivery_expiry_sweep(message_db: "os.PathLike[str]") -> None:
     try:
-        from ai_agent_bridge import _channels as _ch  # noqa: PLC0415 — optional broker bridge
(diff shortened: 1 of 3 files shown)


krisztiankoos:
Opened from the existing `kimi/atlas-6812-identity-flake` worktree after hard_timeout (`needs_finalize`). No new worktree.

Root cause: `scripts/api/preload.py` + `comms_router.py` imported bare `ai_agent_bridge`, which the Monitor API loads inside pytest-xdist. Targeted tests 46 passed. Cross-family review needed at `2a43f38` — **do not spawn a review worktree** until disk migration; comment-only review from an existing seat.

X-Agent: grok-atlas

Issue #6860, opened 2026-08-16
[infra][delegate] read-only mutation guard false-positives on harness runtime state — real deliverable marked failed


krisztiankoos:
Observed 2026-08-16 on task `hetzner-inference-research` (agy, `--mode read-only`, repo root): the worker completed its read-only research and delivered its report (12KB comment posted to the private tracker), yet the task terminalized as `failed` with:

```
last_error = read-only checkout mutation detected: .agent/sessions/a60b06a0-….json, batch_state/fleet-comms/v1/comms.sqlite3-shm, batch_state/fleet-comms/v1/comms.sqlite3-wal
```

All three flagged paths are **gitignored harness runtime state** — the session tracker file and sqlite WAL/SHM side-files created by the worker's own fleet-comms/inbox tooling — not repository content mutations. The guard's intent (catch read-only workers editing the checkout) is right; flagging the harness's own runtime side-effects is a false-positive class that turns successful dispatches into `failed`, breaking settle-loop semantics (orchestrators treat `failed` as needing intervention) and polluting delivery stats.

Fix at the right layer: the mutation detector should evaluate against tracked-file status (`git status --porcelain` on tracked content) or an explicit runtime-state allowlist (`.agent/sessions/`, `batch_state/**` sqlite side-files), not raw filesystem diff. Add a regression test with a fake runtime-state write in a read-only dispatch (mutation-check it, #M-16). Sibling class: #6426 (finalizer misclassification).

krisztiankoos:
Second occurrence (2026-08-16 evening): grok CF-review task `review-6868-grok` (read-only) completed successfully with a full APPROVE verdict, but terminalized `failed` — `read-only checkout mutation detected: .pytest_breadcrumbs/breadcrumb_master.txt, .pytest_cache/*`. The reviewer ran targeted pytest (legitimately, per its brief), and the caches tripped the guard. Add `.pytest_cache/`, `.pytest_breadcrumbs/` to the runtime-state allowlist alongside the earlier `.agent/sessions/` + comms sqlite side-files. This false-failure class is now actively mislabeling successful reviews.

PR #6861, opened 2026-08-16
fix(work): single-flight projection build, startup warmup, and persistent delegate scan cache (#6859)

### LINKED ISSUE #6859: Work: projection endpoint wedges in permanent 504 after Monitor restart — cold build exceeds 5s budget, cache never fills
Found during the final live verification of #6849 (claude-infra, 2026-08-16, Monitor restart ~13:43). The operator saw it live ("the api server is sick").

## Symptom
After `services.sh restart api`, every `GET /api/work/v1/projection` returns 504 `work_projection_timeout` (`budget: warm_target_s=2.0, timeout_s=5.0`). Observed continuously for 3+ minutes / 20+ sequential probes (10s apart) — the board (`work.html`) cannot load at all. `/api/work/v1/health` stays 200 throughout; all other API surfaces fine.

## Measurements (tool-backed, fresh Python process on this host)
Per-section timings via direct calls:
- `fetch_delegate_active`: **11.60s** ← dominates; `_delegate_task_rows` cold-scans **1613** `batch_state/tasks/*.json` (read + derive ≈ 7ms/file; `_TASK_STATE_CACHE` is per-process and empty after restart)
- `fetch_open_issues`: 1.54s · `fetch_open_prs`: 0.87s · `fetch_fleet_reviews`: 0.57s (498 rows, post-#6849 fix) · `fetch_streams_projection`: 0.01s
- `collect_public_sections` (parallel, cold): **20.31s** — notably worse than the sum suggests (thread/subprocess contention; secondary).

So the cold build costs ~12–20s against a 5s request budget, and each request appears to rebuild rather than share one background build — the cache can never become ready under load, so the endpoint is down precisely when a QA operator bounces the Monitor and opens the board (grok-bot's #6849 repro path).

Caveat: the server was externally SIGTERM'd ~2min into observation (see #6830 evidence), so "never warms" vs "warms after minutes" is not fully separated — but a per-request cold rebuild under a 5s fail budget is structurally wrong either way.

## Expected
- A Monitor restart must not take the board down. First page load may show a building/degraded state, but the projection must become servable without any request ever exceeding its budget.
- Likely shapes (implementing lane decides the design): single-flight background projection build + serve-stale-while-revalidate (the streams section already uses exactly this pattern via `schedule_refresh`); warm the projection at lifespan startup; and/or make the delegate scan cheap (cross-process cache / prune-terminal-files-from-scan / index). Fix the architecture, not the budget number.

## Not in scope
The external SIGTERM that killed the restarted server (process-tree kill of agent-shell-spawned services) — that evidence goes to #6830.

Refs #6849 #6842 #6847, grok-bot QA path in #6849.

krisztiankoos:
## Summary
Addresses #6859 — Work projection endpoint wedging in permanent 504 after Monitor restart.

### Root Cause Analysis
1. **Delegate cold scan overhead**: Cold scan across 1615 task state files in `batch_state/tasks/` (~2.84 GB JSON) required 11.6s–13.9s of synchronous I/O and JSON parsing on an empty in-memory cache, consuming nearly 3x the 5.0s projection timeout budget alone.
2. **Cancellation on request timeout**: When a request timed out at 5.0s, the in-flight build was cancelled and abandoned without writing to cache. Under sequential or concurrent requests on a restarted server, each request started another cold build from scratch, timed out, and discarded results, causing the endpoint to wedge in permanent 504.
3. **No single-flight request deduplication**: Multiple concurrent requests spawned concurrent section collectors and thread pools, causing subprocess/thread contention (measured at 20.48s parallel cold build).

### Changes
1. **SQLite Persistent Task Summary Cache (`scripts/api/delegate_router.py`)**:
   - Stores task summary metadata (`mtime`, `task_id`, `status`, `derived_status`, `pid`, `alive`, `started_at`, `duration_s`, `agent`, `model`, `effort`, `cli_version`, `substitution`, `claimed_repo`) in `.task_cache.sqlite3` colocated inside `TASKS_DIR`.
   - On cold server start / process restart, loads the SQLite index in ~2ms; scandir checks file `mtime` against cached values and skips reading/parsing terminal JSON files.
   - Preserves exact zombie reclassification (`_pid_alive` check), repository filtering (`_authoritative_task_repository`), and sorting behavior.
   - Cold scan latency drops from **13,925ms** to **~10.7ms** (1000x+ improvement).

2. **Single-Flight Background Build & Convergence (`scripts/api/work_router.py`)**:
   - Manages in-flight projection builds via `_IN_FLIGHT_BUILDS` so concurrent requests share a single build task.
   - Shields the build task with `asyncio.shield()` so client-side 504 timeouts do not cancel the background build; when finished, it populates `cache_set(key, payload)`.
   - Subsequent retries or concurrent requests immediately converge to 200 OK from warm cache.
   - Falls back to serve-stale-while-revalidate if an expired cache entry is present.
   - Provides `warm_projection_cache()` scheduled non-blocking during API lifespan startup in `scripts/api/main.py`.

### Measurements
- Cold delegate scan on 1615 tasks: **13.92s** (before) -> **10.72ms** (after)
- Cold `collect_public_sections`: **20.48s** (before) -> **4.72s** (after)
- Warm `GET /api/work/v1/projection`: **2.68ms**
- 5 concurrent cold requests to `/api/work/v1/projection`: all completed with 200 OK under 5s budget.

### Testing & Verification
- Unit & regression tests added in `tests/api/test_delegate_active.py` and `tests/test_work_contract.py`.
- Verified mutation testing (#M-16): tests fail when persistent cache loading or single-flight deduplication is removed.
- All 73 affected tests pass in pytest.
- Ruff linter clean.

Closes #6859

### DIFF
--- a/scripts/api/delegate_router.py
+++ b/scripts/api/delegate_router.py
@@ -5,8 +5,10 @@ from __future__ import annotations
 import asyncio
 import json
 import os
+import sqlite3
 import time
 from datetime import UTC, datetime
+from pathlib import Path
 from typing import Any, Literal
 
 from fastapi import APIRouter, HTTPException, Query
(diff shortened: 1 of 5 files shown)


krisztiankoos:
CROSS-FAMILY REVIEW OF RECORD — reviewer: claude-infra (Anthropic/Fable, inline seat), author: agy (Google). VERDICT: APPROVE.

Reviewed the full diff at head 3dd6449d6a against #6859:

1. **Single-flight + shield**: correct pattern — `asyncio.shield` keeps the build alive past client timeouts, `cache_set` on completion, and the `finally: _IN_FLIGHT_BUILDS.pop` guarantees a failed build never wedges the key. Concurrent requests attach to one build (mutation-checked: 8→1).
2. **Stale-serve on timeout**: serves the aged cache with honest `cache_age_s` instead of 504 — matches the issue's Expected exactly.
3. **Lifespan warmup**: scheduled non-blocking inside the running loop, exception-guarded — a restart now starts the cold build immediately instead of waiting for the first victim request.
4. **SQLite persistent summary index**: placed at `batch_state/tasks/.task_cache.sqlite3` — gitignored runtime dir, name invisible to the `.json` scanner; single-writer (the server); mtime-diffing keeps zombie/pid liveness exact. 13.9s → 10.7ms cold is the measured result and the mutation check proves the load path is real (51→0 reparse assertion).
5. Numbers: cold collect 4.72s now fits the 5s budget once, single-flight makes it once-only, and warmup makes even that invisible. Budget untouched — architecture fixed, not the number.

Non-blocking note: an in-process consumer of `active_delegate_tasks` inside a read-only dispatch would write the cache db and trip the #6860 false-positive class — #6860's runtime-state allowlist covers it; no action here.

Residual: final user-level verification (restart the live server, cold-probe the endpoint and the board) is the merge owner's follow-through after this lands. Arming auto-merge.

PR #6873, opened 2026-08-16
fix(bridge): align inbox banner with live-pending store (#6864)

### LINKED ISSUE #6864: [infra][comms] Legacy inbox banner: wrong count (20 vs 8), stale age, recommends retired 'inbox run' in authority mode
Observed 2026-08-16 (claude-infra): every bridge invocation prints
`⚠️ grok-infra has 20 pending deliveries (oldest 2d17h→2d22h). Run 'ab inbox run grok-infra' to drain.` — three defects in one banner:

1. **Count/age wrong:** `inbox show grok-infra` reports pending: **8**, oldest **15h19m** — the banner counts something else (retired/acked legacy rows?) and its age only ever grows, which panicked an operator status report ("3-day backlog") over what is mostly test traffic (oldest preview: `[test-slot/deec2] user → "Hello slot"`).
2. **Recommends a retired command:** `inbox run` in authority mode refuses with "legacy `inbox run` writes are retired in authority mode; use `python -m scripts.fleet_comms channel ...`" — but no fleet_comms subcommand actually drains agent deliveries; the real consumer is the agent's driver session (drive-epic §0a). The banner should say that, or nothing.
3. **`ab` alias in the hint** — bare `ab` resolves to ApacheBench per project guardrails; the printed hint violates the project's own rule.

Fix at the right layer: banner counts from the authority store (same query as `inbox show`), hint text names the actual consumption path, drop the `ab` alias. Also consider expiring/acking `test-slot/*` deliveries so test traffic doesn't age into scary numbers.

Grok lane itself verified healthy today (`GROK-OK grok-4.5` via ask-grok) — the handoff's "grok 402" note is stale; this banner was the main thing keeping the scare alive.

krisztiankoos:
## Summary

Fixes #6864 — the legacy inbox backlog banner reported a wrong pending count, ever-growing age, and a retired `ab inbox run` drain hint.

**Root cause:** the banner counted *all* `pending` deliveries, while `inbox show` runs `expire_stale_deliveries()` first and then counts. On one invocation that printed e.g. 20 then showed 8; ages kept growing on TTL-elapsed rows the banner still counted.

**Fix:**
- Shared authority query `live_pending_by_agent` / `live_pending_for_agent` — within-TTL pending only (same set show would report after expire), used by both banner and show
- Banner hint: drain in the live driver session (drive-epic §0a); no retired `inbox run`, no bare `ab`
- Operational: expired 8 live `test-*` harness pending deliveries for grok-infra via `mark_delivery(..., "expired")` so test traffic stops aging into scary numbers

## Test plan

- [x] `pytest` banner / live-pending fixtures:
  - `test_backlog_banner_triggers_for_old_pending_delivery`
  - `test_backlog_banner_count_matches_inbox_show_excluding_ttl_elapsed` (mutation: seed 1 live + 1 past-TTL → both surfaces report 1)
  - related pending/expire bridge tests (21 passed)
- [x] ruff clean on touched files
- [x] live: `inbox show grok-infra` pending 0; no grok-infra banner on `channel list`

## Residual

- Fleet-comms `authority_deliveries` backlog for other agents is a separate surface (`fleet_comms backlog`); this PR only fixes the legacy channel-delivery banner that `ai_agent_bridge` prints.
- No auto-merge (dispatch instruction).

### DIFF
--- a/scripts/ai_agent_bridge/_channels.py
+++ b/scripts/ai_agent_bridge/_channels.py
@@ -1881,30 +1881,106 @@ def pending_deliveries_for(agent: str) -> list[dict[str, Any]]:
         conn.close()
 
 
-def bridge_pending_summary(now: str | None = None) -> dict[str, dict[str, float | int]]:
-    """Return per-agent pending counts and oldest pending age in hours."""
-    current = _parse_iso(now) if now else datetime.now(UTC)
+def live_pending_by_agent(now: str | None = None) -> list[dict[str, Any]]:
+    """Return per-agent live pending counts (authority query for banner + show).
+
+    Counts only ``status='pending'`` rows that would *remain* after
+    ``expire_stale_deliveries`` — i.e. still within their effective TTL.
+    Read-only: never mutates rows. Shared by the CLI backlog banner and
+    ``inbox show`` so the two surfaces cannot disagree (#6864).
+    """
+    current = now or _now_iso()
     conn = get_db()
     try:
         rows = conn.execute(
-            """
-            SELECT d.to_agent, COUNT(*) AS count, MIN(cm.created_at) AS oldest_created_at
+            f"""
+            SELECT d.to_agent AS to_agent,
+                   COUNT(*) AS count,
+                   MIN(cm.created_at) AS oldest_created_at
             FROM deliveries d
             JOIN channel_messages cm ON cm.message_id = d.message_id
+            JOIN channels c ON c.name = cm.channel
             WHERE d.status = 'pending'
+              AND (
+                    julianday(:now) - julianday(cm.created_at)
+                  ) * 24.0 <= {_effective_ttl_hours_sql()}
             GROUP BY d.to_agent
             ORDER BY d.to_agent ASC
-            """
+            """,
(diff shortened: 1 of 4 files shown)


krisztiankoos:
CROSS-FAMILY REVIEW OF RECORD — reviewer: claude-infra (Anthropic/Fable, inline seat), author: grok-build (xAI). VERDICT: APPROVE.

The structural root cause is exactly right and fixed at the right layer: banner counted all `status='pending'` rows while `show` expired TTL-elapsed ones first — now both surfaces share one read-only live-pending authority query with the TTL predicate in SQL, so they cannot disagree by construction. Hint text no longer names the retired path or bare `ab`; the eight `test-*` junk deliveries were expired operationally (banner now silent, verified). Mutation fixture covers the disagreement case; 21 tests green, ruff clean.

Non-blocking nits for the record: (1) the added `JOIN channels` is an INNER join — a delivery whose channel row vanished would silently drop from the count; if orphaned deliveries are possible, a LEFT JOIN + default TTL would count them (worth a follow-up test either way). (2) `_effective_ttl_hours_sql()` interpolated via f-string is fine while it stays a constant internal expression — keep dynamic values parameterized as done with `:action_ttl`.

Arming auto-merge.

PR #6874, opened 2026-08-16
fix(tests): remove checkout-venv dependency from launcher contract suite (#6858)

### LINKED ISSUE #6858: [infra][tests][flake] test_launcher_contract.py: 7 driver-termination/lease tests fail locally at clean base (child-started sentinel)
Independently reproduced by two agents on 2026-08-16 (claude lane in a dispatch worktree; kimi K3 during CF review of PR #6857, re-run at clean base `0fd98b8eaf` in the primary tree):

- `test_driver_termination_forwards_signal_and_closes_exact_lease[*]` (parametrized) and `test_real_store_driver_close_successor_and_expired_recovery` — 7 failures total — fail on a subprocess `child-started` sentinel that never arrives.
- Same failures at head and at base ⇒ environmental/flaky, not introduced by any current PR.
- One prior report attributed it to the missing worktree-local `.venv` (project policy), but kimi reproduced at base in the PRIMARY checkout, so venv absence is not the whole story.

Why it matters: these guard the driver-termination/lease contract that the drive-script onboarding relies on; a permanently-red-locally suite trains agents to ignore it (CF reviewers must currently stash-diff against base to prove pre-existence — wasted effort each review).

Ask: root-cause the sentinel race (timing? signal delivery under sandbox? tmpdir?), then either fix the harness or mark with a deterministic skip-with-reason so local red means something again. Mutation-check any fix (#M-16).

krisztiankoos:
Closes #6858.

## What the flake actually was

Not a race. All 7 locally-red tests (`test_dry_run_does_not_require_a_provider_binary`, `test_canary_failure_closes_lease_and_refuses_driver_launch`, `test_driver_normal_exit_closes_with_bounded_idempotent_retry`, `test_driver_termination_forwards_signal_and_closes_exact_lease[2-130|15-143|1-129]`, `test_real_store_driver_close_successor_and_expired_recovery`) are the deterministic signature of a checkout with no `.venv/bin/python` — which is every dispatch worktree, by the shared-interpreter policy. Measured chain:

1. The exit/canary fixtures' stub python intercepts the two supervisor module calls but falls back to `exec REPO/.venv/bin/python` for everything else. The claim helper's heredoc capsule-parse step hits that fallback → exit 127 → the driver dies **before** spawning the provider → the `child-started` sentinel never appears (the reported symptom).
2. The real-store test's wrapper python execs the same missing path for all calls → the supervisor claim fails → launch refused (rc 1 ≠ 0).
3. `start-claude.sh` preflight pins `CLAUDE_PROFILE_RESOLVER_PYTHON=$LC_ROOT/.venv/bin/python` (scripts/launchers/claude.sh:9) and exits 2 even under `LAUNCHER_DRY_RUN=1` → the dry-run test fails.

Running the identical file at the venv-present primary: **50/50 green**. Closed-stdin and timing hypotheses were tested and eliminated. CI is green because CI builds `.venv` at the checkout root.

## Fix (harness layer, no sleeps, no blanket skips)

- Fixture stubs/wrapper now fall back to `sys.executable` (the interpreter running pytest — always the shared project python), so the six fixture tests genuinely run in any checkout, worktrees included.
- The dry-run test exercises the real adapter, whose checkout-venv requirement is the product contract; it gets a deterministic `skipif` gated on the **detected** condition (missing `REPO/.venv/bin/python`) with the policy reason spelled out.
- Sentinel wait hardened: 10 s monotonic deadline (was a fixed 2 s loop), and on failure it reports the launcher exit code + captured output instead of a bare `assert` — with this message the original reports would have been self-diagnosing.

## Proof (#M-4 / #M-16)

- Pre-fix worktree run: `7 failed, 43 passed` — exactly the issue's set (reverting the fix reproduces it).
- Post-fix: **10 consecutive** full-file runs in the previously-failing worktree, each `49 passed, 1 skipped`.
- Mutation checks: non-executable provider trips the new diagnostic (`provider child never started (launcher rc=126)` + quoted stderr); the skip fires only where the venv is absent (runs and passes at the primary).
- `ruff check`: clean.

## Follow-up (out of my owned paths)

`scripts/launchers/claude.sh` could fall back to the durable-helper root's venv (`LC_DURABLE_HELPER_ROOT`) so `start-claude.sh --help`-beyond flows work from worktrees; noted on the issue for a future in-lane task.


krisztiankoos:
CROSS-FAMILY REVIEW OF RECORD — reviewer: grok-4.6 (xAI, native CLI seat), author: claude-fable-5 (Anthropic). Relayed by claude-infra (merge owner; reviewer's harness blocks external posting). VERDICT: APPROVE.

son observed: checkout `.venv` / shared-interpreter policy (#6858). Matches the PR claim.

6. **Sibling residual (same anti-pattern; out of #6858 surface but should be swept):**
   ```
   tests/test_start_codex_driver.py:55:exec {os.fspath(REPO / ".venv" / "bin" / "python")!r} "$@"
   tests/orchestration/test_thread_restart_e2e.py:186:exec {os.fspath(REPO_ROOT / ".venv" / "bin" / "python")!r} "$@"
   ```
   Same checkout-relative fallback class. Driver already queued a worktree-hostile sweep. Residual debt, **not** a change request on this head.

7. **Autopsy index entry** is accurate (env-dependent red misfiled as race; prevention template). No issue.

### Non-blocking residual
- Product follow-up (claude adapter fallback via primary/durable-helper venv for real worktree dry-run) remains out of scope; skip is honest until then.
- No evidence for the original race framing.

---

**VERDICT: APPROVE**

Root-cause chain is tool-backed; fix is correctly layered; suite is green in the venv-less environment that produced the original red. Sibling stub debt is follow-up, not blocking.



```bash
gh pr comment 6874 --body-file <path-to-the-review-body-above>
```

PR #6837, opened 2026-08-15
fix(comms): restore ask-gemma ACP route, actionable hermes-seat failure (#6805)

krisztiankoos:
Residual from #6805. Root cause (primary-checkout acpx clobber) was fixed there and ask-glm verified working; this PR restores the two repo-fixable reviewer-seat routes that stayed dead for reasons of their own.

## (a) ask-hermes / DeepSeek seat — actionable error + fail at admission

Hermes is genuinely host-level (operator-owned platform + credentials under `~/.hermes/`; the machine-local install recipe is deliberately untracked per the 2026-07-05 OPSEC policy), so provisioning is **not** automatable from project tooling. Shipped instead: actionable error + reroute path + runbook provisioning steps.

- `acpx.py::_missing_binary_message` — every missing provider binary now names the install/verify step, the owning runbook section, and (for hermes) the documented fallback substitutions from `agents_extensions/shared/rules/model-assignment.md`: `opencode run --model deepseek-direct/deepseek-v4-flash --variant high` for one-shot review, `delegate.py dispatch --agent deepseek` for tool-heavy work.
- `acpx.py::probe_participant_reachability` — PATH-only preflight (no subprocess) called from `_acp_compat._run_compat_ask_impl` **before** an authority job is enqueued, so callers degrade at admission instead of dying mid-review. The adapter's compatibility probe still enforces the full contract before spawn.
- `docs/runbooks/agent-seat-onboarding.md` — new "Reviewer-seat transport recovery" section: symptom→meaning table, Hermes provisioning + verification (`hermes --version` → `Hermes Agent v<semver>`, required flags), documented fallbacks.

**Before** (`ask-hermes "Reply with exactly: OK" --task-id probe-6805-before-hermes --from kimi`):
```
AcpxDeepSeekShadowAdapter: hermes binary not found on PATH; required for compatibility contract 'text-oneshot-isolated-v1'
```

**After** (`ask-hermes ... --task-id probe-6805-after-hermes --from kimi`, exit 1, no quota spent — refused at admission):
```
[...]

PR #6878, opened 2026-08-16
fix(bridge): DeepSeek ask route via opencode ACP + review-ask outcome validation (#6805)

krisztiankoos:
Residual from #6805 (the 2026-08-16 comment): the DeepSeek seat had a documented standing route (`opencode run --model deepseek-direct/deepseek-v4-flash`) but no working bridge ask path, and transport-ok masked garbage review replies.

## Decision: outcome (a) — enable the ACP route

Hermes was permanently removed from the host (operator order 2026-08-16), so the seat's old Hermes text-agent route could never work again — outcome (b) would have documented a dead seat instead of fixing a live one. The standing opencode route supports the confined native shape already proven by the GLM and Gemma seats (#6837), so the seat now mirrors that pattern exactly:

- `AcpxDeepSeekShadowAdapter` — native `opencode acp --pure` with the deny-all OpenCode config, pinned to `deepseek-direct/deepseek-v4-flash` @ high (Flash is the review-ladder rung; Pro stays hard-implement-only per the 2026-08-13 GO). The China-egress CI refusal is retained (`DEEPSEEK_FIRST_PARTY_FORBIDDEN`).
- Bridge gains **`ask-deepseek`**; `ask-hermes` stays as an alias of the same seat; the generic `ask-opencode` route stays retired (fails closed with a clear error).
- The admission-time reachability probe and the model-catalog endpoint now key the DeepSeek participant on the `opencode` binary, and reviewer-seat guidance (`model-assignment.md`, seat-onboarding runbook, runtime guide) no longer points at the dead hermes path.

## Outcome validation for review asks (the issue's core residual)

For `--type review` (what drivers actually pass — it previously set no review intent at all) and `--review` asks, a transport-ok reply must contain:

1. a `VERDICT` token, and
2. at least one finding/evidence line or concrete artifact reference (`path[:line]`, `#PR`, or commit SHA),

else the authority job terminalizes **`failed:non_evidentiary`** (`postprocess` phase, non-retryable in place) instead of a silent `replied` — the reply body is preserved on the receipt for forensics. Cheap and deterministic; no LLM judge in the transport layer.

## Evidence (#M-4), mutation-checked (#M-16)

Live probes on this branch:

```
$ ask-deepseek "Reply with exactly: OK deepseek-v4-flash" --task-id probe-6805-deepseek-ask --from kimi
OK deepseek-v4-flash
deprecated ask-deepseek: ACP transport; outcome=ok
```

```
$ ask-deepseek - --task-id probe-6805-deepseek-review --type review --from kimi
VERDICT: CORRECT
Evidence: in `- cp "$src" "$dst"` → `+ install -m 0644 "$src" "$dst"`, the destination mode is now forced to 0644 …
deprecated ask-deepseek: ACP transport; outcome=ok
```

Mutation checks: disabling the verdict gate, disabling the evidence gate, or reverting the provider-binary mapping to `hermes` each fails the new tests.

Tests: 1146 passed across `tests/ai_agent_bridge/`, `tests/agent_runtime/`, `tests/test_agent_runtime.py`, reviewer-resolver, and model-catalog suites; the only 3 failures are pre-existing sparse-checkout/venv environment artifacts (verified identical on clean HEAD via stash). `ruff check` clean. `deploy_prompts.sh --dry-run` errors are all pre-existing sparse-checkout `ORCH_TRACK_NOT_ACTIVE` rows.

Known cosmetic non-issue: `dispatch telemetry for acpx-deepseek-shadow could not resolve cli_version … recording 'unknown'` — the generic best-effort telemetry helper has no version probe for any `acpx-*-shadow` seat (glm/gemma behave identically); telemetry records `unknown` and the ask proceeds.

No auto-merge armed.


krisztiankoos:
CROSS-FAMILY REVIEW OF RECORD — reviewer: claude-infra (Anthropic/Fable, inline seat), author: kimi K3 (Moonshot). VERDICT: APPROVE.

1. **DeepSeek seat restoration** — right call on outcome (a): mirrors the proven GLM confinement shape (`opencode acp --pure`, deny-all, pinned `deepseek-direct/deepseek-v4-flash`, China-egress CI refusal retained), `ask-deepseek` + legacy alias, generic `ask-opencode` stays failed-closed. Live probe quoted (`OK deepseek-v4-flash`, outcome=ok). Runbook/rules text now points nowhere near the dead hermes path.
2. **Non-evidentiary gate** — deterministic and correctly calibrated at the false-positive edge I probed: a clean "VERDICT: APPROVE — no findings" passes (the explicit findings statement grounds it), while a bare ungrounded verdict fails — which is the right policy: a review that cites nothing is not a review. `retryable: False` with the fresh-task-id guidance is exactly right (idempotent replay would repeat the garbage). Both behaviors mutation-checked; live review-probe shows no false positive on a real graded reply.
3. Rebase over the concurrent grok-4.6 rotation preserved main's pins (checked `model-assignment.md` on the branch). 1066 tests green with the 3 known env-class failures verified pre-existing.

Arming auto-merge.

Issue #6876, opened 2026-08-16
[4387][atlas] First protocol-tracked job on atlas-runner (post-#6867)


krisztiankoos:
Campaign issue for the first `atlas-job.v1` run after #6867.

**Outcome:** one tracked job on **atlas-runner** (not hramatka) with a fail-closed result receipt. Smoke first (`--limit 5`), then a real residual pass.

**Do not** re-run the 562-slug missing-translation list as-is: last workdir run died on missing `yaml` and still reported systemd success (empty summary). Protocol close must not treat that as succeeded.

**Non-goals:** no pointer flip, no hramatka batch, no #6809 undraft.

Parent: #4387. Protocol: #6867 (`7afed18`).

krisztiankoos:
AGY free-disk preflight landed as PR #6879 (`fd38e0a003`). CF will be a non-Gemini family.

krisztiankoos:
Kimi venv fail-closed PR: #6881 (`125d30c792`). AGY CF in flight. DeepSeek CF of #6879 retried after abort.

krisztiankoos:
## Protocol smoke `6876-smoke-limit5` (Cursor)

Tracked job on **atlas-runner** only (`pointer_write: false`). Plan: `full-catalog` + `--limit 5` (not the stale 562-slug residual).

| Field | Value |
| --- | --- |
| state | **succeeded** |
| delivery | **ok** |
| targets | 5 |
| changed | 5 |
| filled_translation | 0 |
| consecutive_misses | 0 |
| circuit_breaker_tripped | false |
| categorical_binning | ENRICHED 5 |
| backup.ok | true |
| unit | atlas-job-6876-smoke-limit5.service (inactive) |
| journal | terminal after status reconcile → close |

Receipt: `batch_state/atlas-jobs/receipts/6876-smoke-limit5.json` (schema-capped; no hostnames/IPs/secrets).

Smoke also surfaced and fixed: fresh workdir `data/` overlay creation; shared project interpreter for restic gate receipts from dispatch worktrees; git receipt `delivery` stamped before write.

No pointer flip. Remote workdir retained. PR opened from `cursor/6876-protocol-smoke-cursor` (not merged).

krisztiankoos:
GLM security review: two attempts, no verdict. First pass cut by read-only guard; r2 `glm-5.3 rate limited`. Not looping GLM. #6881 AGY APPROVE + auto-merge armed.

[...]

PR #6881, opened 2026-08-16
fix(runner): fail closed when runner venv lacks yaml/jsonschema (#6876)

krisztiankoos:
## Summary

Campaign #6876. The last atlas-runner job died with `ModuleNotFoundError: yaml` **after** systemd-run had already reported success, because `launch_reenrich_class_b.sh` only checked that `$REPO/.venv/bin/python` existed — never that it could import the driver's hard deps.

## Changes

- `scripts/lexicon/runner/launch_reenrich_class_b.sh`
  - New preflight before any filesystem/systemd side effect: resolves the runner interpreter — `$REPO/.venv/bin/python` preferred, `$CODE_ROOT/.venv/bin/python` fallback, `ATLAS_RE_ENRICH_RUNNER_PYTHON` override — then runs `$RUNNER_PYTHON -c 'import yaml, jsonschema'`.
  - On failure: prints a one-line install recipe (`<py> -m pip install pyyaml jsonschema`) and exits 1. No apt-get, no silent success.
  - The wrapped driver command (`run_cmd`) now uses the preflight-resolved `$RUNNER_PYTHON` instead of a hardcoded `$REPO/.venv/bin/python`, so the check and the launch can never diverge.
- `launch_reenrich_class_b_remote.sh` — **no change needed**: it invokes the launcher over ssh under `set -euo pipefail`, so the remote exit 1 propagates and the wrapper already fails closed.
- `tests/test_launch_reenrich_venv_preflight.py` (new, 7 tests): drives the real launcher against a fully stubbed tmp tree (stub venv pythons, stub driver/data) — fail-closed on missing imports, fail-closed on missing venv, happy path, `$CODE_ROOT` fallback, preference order, override, and a source contract tying `run_cmd` to `$RUNNER_PYTHON`.

## Verification

- `.venv/bin/python -m pytest tests/test_launch_reenrich_venv_preflight.py tests/test_launch_reenrich_target_detection.py -v` → **27 passed**
- `ruff check` on the new test → clean; `bash -n` on the launcher → clean

## Out of scope (per dispatch)

No changes to `atlas_job.py` (AGY owns free-disk) or `atlas_jobs_router.py` (GLM reviewing). No merges, no pointer flips.

Refs #6876

### DIFF
--- a/scripts/lexicon/runner/launch_reenrich_class_b.sh
+++ b/scripts/lexicon/runner/launch_reenrich_class_b.sh
@@ -31,7 +31,7 @@
 #
 # Env overrides:
 #   ATLAS_RUN_ROOT, ATLAS_REPO, ATLAS_RE_ENRICH_WORK_DIR,
-#   ATLAS_RE_ENRICH_CODE_ROOT,
+#   ATLAS_RE_ENRICH_CODE_ROOT, ATLAS_RE_ENRICH_RUNNER_PYTHON,
 #   ATLAS_RE_ENRICH_UNIT, ATLAS_RE_ENRICH_DRIVER, ATLAS_RE_ENRICH_SLUGS_FILE,
 #   ATLAS_SOURCES_DB, ATLAS_KAIKKI_JSON, ATLAS_LIVE_MANIFEST
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


krisztiankoos:
## Cross-family review (AGY / Gemini) — APPROVE

Exact head: `125d30c792c8e7f3f8d3d01840683809f8013ea2`. Author is Kimi. Wrapper `review-6881-agy` rc=0; file marked `failed` only for unrelated primary-checkout mutations.

Preflight exits 1 before systemd-run when yaml/jsonschema missing; recipe printed; no apt-get; 7 tests; remote wrapper stays fail-closed. Arming auto-merge; waits on CI Gate.

PR #6926, opened 2026-08-17
fix(acp): unwedge ask-cursor replay guard (#6916)

### LINKED ISSUE #6916: [infra][acp] ask-cursor wedged: every fresh request fails with 'duplicate terminal response replay detected for the same request id'
Observed live 2026-08-17 (~08:5x), twice in a row with DISTINCT `--task-id` values (`analyze-776`, `analyze-776-r2`):

```
deprecated ask-cursor: ACP transport; outcome=error
duplicate terminal response replay detected for the same request id
```

Every new `ask-cursor` request fails the same way, so this is not a caller-side id collision — the ACP controller state for the cursor route appears wedged on a stale terminal response that gets replayed against fresh request ids. The cursor DELEGATE lane is unaffected (multiple successful dispatches same day: infra-6890, infra-6880-doc, infra-6870-r2, infra-492 increments).

Asks:
- Root-cause the replay-guard state for the cursor ACP route (what terminal receipt is being replayed, why it matches fresh request ids) and add a recovery path that does not require manual state surgery.
- The error should name the offending request/receipt id so the operator can see WHICH stale response is stuck — today it is unactionable.

Workaround in use: route asks meant for cursor through `delegate.py dispatch --agent cursor --mode read-only` (noted substitution).

Refs #6915 (bridge process-path defects found the same morning; different layer — that one is routing/ack, this one is ACP replay state).

krisztiankoos:
Fixes #6916.

Authored by the grok lane (dispatch `infra-6916-acp-replay-wedge`); driver finalized push/PR because the grok harness could not run network git this session — code is entirely the lane's commit `eea16b0c2f`.

## Root cause (from the lane's evidenced analysis)
The ACP replay guard treated any repeated terminal response id as a replay, but Cursor's route reuses small integer request ids across methods (`session/request_permission` receipts colliding with later `session/prompt` requests) — so one stale receipt matched every fresh request and permanently wedged `ask-cursor` with `duplicate terminal response replay detected`.

## Fix
- Replay detection is now per **request-id generation**: a fresh `method+id` request supersedes the prior receipt for that id (`terminal_generations` lifecycle in `scripts/agent_runtime/adapters/acpx.py`).
- A true double-result with no intervening request still fails closed, and the error now names the offending id and method (the issue's second ask).
- Recovery path without manual state surgery: a completed, non-cancelled `stopReason` on ACPX exit-5 is accepted.

## Evidence (lane-run, quoted from the task result)
- `ruff` clean; `pytest tests/agent_runtime/test_acpx_adapter.py tests/ai_agent_bridge/test_ask_contract.py` → **213 passed**.
- Mutation check: removing `terminal_generations.discard(event_id)` fails both new guard tests with the exact wedge error; restore → pass.
- Live recovery probe (driver-run post-push): see PR comment.

X-Agent: grok/infra-6916-acp-replay-wedge

### DIFF
--- a/scripts/agent_runtime/adapters/acpx.py
+++ b/scripts/agent_runtime/adapters/acpx.py
@@ -740,6 +740,26 @@ _STOP_REASONS = frozenset(
 )
 _STOP_REASON_CANCELLED = "cancelled"
 _MISSING_STOP_REASON = object()
+# acpx EXIT_CODES.PERMISSION_DENIED: deny-all refused a tool/permission
+# request. The prompt can still complete with a valid stopReason.
+ACPX_EXIT_PERMISSION_DENIED = 5
+
+
+def _is_jsonrpc_id(value: object) -> bool:
+    """JSON-RPC request ids are string | number; JSON ``true``/``false`` are not."""
+    return value is not None and not isinstance(value, bool) and isinstance(value, (str, int, float))
+
+
+def _jsonrpc_id_text(value: object) -> str:
+    return json.dumps(value, separators=(",", ":"))
+
+
+def _duplicate_replay_reason(request_id: object, method: str | None) -> str:
+    """Name the colliding request so the operator can see which receipt is stuck."""
+    labelled = f"request id {_jsonrpc_id_text(request_id)}"
+    if method:
+        labelled = f"{labelled} ({method})"
+    return f"duplicate terminal response replay detected for {labelled}"
 
 _USAGE_TOTAL_FIELDS = ("total_tokens", "totalTokens")
 _USAGE_INPUT_FIELDS = ("input_tokens", "inputTokens")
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Live probe result (driver, post-push): `ask-cursor 'PONG-6916'` → exact reply `PONG-6916`, outcome=ok. Scope honesty (#M-4a): this probe ran against the PRIMARY checkout's unfixed adapter, so it proves the cursor route is healthy again (the wedging receipt has been superseded by natural state turnover) — it does NOT prove the fix caused the recovery. What proves the fix: the two new guard tests reproduce the exact wedge error on mutation (removing the generation-discard) and pass at head, i.e. the class of wedge #6916 describes becomes impossible rather than merely cleared.

krisztiankoos:

The reviewer walked the interleavings explicitly: fresh method+id request supersedes the prior receipt (accept), true double-terminal with no intervening request still fails closed with the id+method-naming diagnostic, and the exit-5 acceptance preserves deny-all semantics (permission already denied at protocol level; any crash, other exit code, missing/cancelled stopReason, or protocol error still fails closed). Tests confirmed as genuine guards, including the seat-agnostic parameterization across all four shadow adapters — so the sibling-wedge question from the issue is covered by contract, not just for cursor.

Review gate passed → arming auto-merge; merges on green blocking CI. Note per #M-19: the bridge/ask path executes from the checkout per-invocation, so no service bounce is needed for this one — the fix is live for CLI callers on the next pull.

PR #6927, opened 2026-08-17
fix(delegate): refuse silent primary binding from sibling cwd (#6900)

### LINKED ISSUE #6900: [infra][delegate] dispatch cannot target a sibling repo worktree cleanly — --worktree pins the primary repo, --cwd+--worktree conflict, misbinding is silent
Hit live 2026-08-16 while dispatching a private-repo (sibling checkout) task:

1. `dispatch --worktree --cwd <sibling-repo-root>` → hard error `--cwd cannot be combined with --worktree` (correct refusal, fine).
2. `dispatch --worktree` invoked WITH shell cwd inside the sibling repo → **silently created the worktree in the primary repo** (`_REPO_ROOT` is resolved from `delegate.py`'s own path, delegate.py:131), so the worker would have run in the wrong repository while its brief said otherwise. Caught and cancelled before the worker acted.
3. Working flow (undocumented): manually `git worktree add` in the sibling repo, then `dispatch --mode workspace-write --cwd <that-worktree>` with no `--worktree` — the verified-added-worktree path accepts it.

Asks (root-cause layer, not a wrapper):
- Make the repo binding EXPLICIT and verified: when the invocation cwd (or a new `--repo` flag) resolves to a different git root than `_REPO_ROOT`, either support it first-class (derive the dispatch worktree under THAT repo) or refuse loudly ("dispatch targets the primary; you are elsewhere") instead of silently binding the primary.
- Document the sibling-repo flow (manual worktree + `--cwd`) in the delegate runbook until/unless first-class support lands.

Evidence: task `private-492-private-increment` (cancelled, worktree reaped) vs `private-492-private-increment-r2` (correct, via manual worktree + --cwd).

krisztiankoos:
Fixes #6900.

Authored by the grok lane (dispatch `infra-6900-delegate-repo-binding`, commit `e7a337560a`); driver finalized push/PR (grok harness cannot run network git).

## Summary
`--worktree` / `--branch` always derive the dispatch worktree under `_REPO_ROOT`. Invoking `dispatch --worktree` with the shell cwd inside a sibling git root used to silently create the worktree in the primary repo (#6900's live repro).

**Shape chosen: (b) loud refusal**, not first-class `--repo`: fetch, `git worktree add`, the reaper, sparse-checkout, and data-symlink provisioning all assume `_REPO_ROOT`, so first-class multi-repo support would ripple through all of them. The already-working manual-worktree + `--cwd` sibling flow is now documented in `docs/runbooks/agent-seat-onboarding.md`.

`--branch` had the same cross-repo blindness — the new guard covers both entry points; explicit `--cwd` into a verified added worktree remains allowed.

## Evidence (lane-run, quoted from the task result)
- Before/after probe: silent `derived_under_primary=True` from a sibling cwd becomes `worktree_guard_refuses=True` with the "different git root" message; explicit `--cwd` still allowed.
- `pytest tests/test_delegate.py` → **262 passed**; ruff clean.
- Mutation check: forcing the new root-resolution helper to return `_REPO_ROOT` re-silences the bind and fails the guard tests; restore → refuses again.

X-Agent: grok/infra-6900-delegate-repo-binding

### DIFF
--- a/docs/runbooks/agent-seat-onboarding.md
+++ b/docs/runbooks/agent-seat-onboarding.md
@@ -202,6 +202,38 @@ active environment; the merged-PR worktree reaper disposes of it safely.
 Workers implement inside `.worktrees/dispatch/<agent>/<task>/`. They do not
 become a second coordination plane.
 
+#### Sibling-repo dispatch
+
+`--worktree` and `--branch` always bind the Learn Ukrainian primary checkout
+that owns `scripts/delegate.py` (`_REPO_ROOT`). They do **not** follow the
+shell cwd. Invoking `dispatch --worktree` from a sibling repository (for
+example a private checkout next to this one) is refused: the worktree would
+otherwise be created under the primary while the brief still described the
+sibling.
+
+`--cwd` cannot be combined with `--worktree` or `--branch`. That refusal is
+intentional. The supported sibling-repo flow is:
+
+```bash
+# In the sibling repository — not via delegate --worktree:
+git -C /path/to/sibling worktree add .worktrees/dispatch/<agent>/<task> <base-or-branch>
+
+# Invoke this repo's delegate against that worktree (no --worktree).
+# Derive LU_PRIMARY from the Learn Ukrainian checkout, never from the sibling.
+LU_PRIMARY="$(dirname "$(git -C /path/to/learn-ukrainian rev-parse --path-format=absolute --git-common-dir)")"
+"$LU_PRIMARY/.venv/bin/python" "$LU_PRIMARY/scripts/delegate.py" dispatch \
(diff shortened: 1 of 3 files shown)


krisztiankoos:

Confirmed on the hard edge I asked about: primary linked worktrees do NOT trip the guard (git-common-dir resolution, with an explicit test for invocation from a primary worktree), the three rejection triggers and the two allowed flows (explicit --cwd added worktree; non-git cwd) each carry their own test, end-to-end dispatch tests verify exit 2 + no spurious worktree/state creation, and the mutation check is genuine. Refusal message names the supported sibling flow, which is also now in the onboarding runbook.

Review gate passed → arming auto-merge; merges on green blocking CI. No service bounce needed (#M-19: delegate is per-invocation CLI).

PR #6928, opened 2026-08-17
fix(bridge): route process/ask-gemini via the live ACP participant registry

### LINKED ISSUE #6894: [infra][transport] ask-gemini default model slug is stale — fails against agy ACP pin (gemini-3.7-flash-high)
Observed live 2026-08-16 (~22:5x) firing a review ask:

```
agent_runtime.runner.InterAgentTransportError: ACP participant 'agy' only supports its
registered model pin 'gemini-3.7-flash-high'; got 'gemini-3-flash-preview'
(scripts/agent_runtime/runner.py:3158, resolve_inter_agent_route)
```

`ask-gemini` without `--model` defaults to `gemini-3-flash-preview`, which the AGY route
rejects at `resolve_inter_agent_route`. Workaround that worked: pass
`--model gemini-3.7-flash-high` explicitly.

Root-cause layer to check: the bridge's legacy Gemini default slug vs the single source of
the AGY pin — the default should derive from the registered participant pin (or the route
should map legacy slugs), not carry an independent hardcoded model name that goes stale on
every rotation. Sibling sweep: check other ask-* seats for independently hardcoded default
slugs that can drift from their ACP pins the same way.

Non-urgent (explicit --model unblocks), but every bare `ask-gemini` is currently broken.

### LINKED ISSUE #6915: [infra][bridge] generic 'process <id>' ignores the recipient seat, shells to the retired gemini CLI, and consumes the message on failure
Observed live 2026-08-17 draining bridge message #776 (a qa-engineer advisor note addressed to `cursor`):

```
🤖 Processing with Gemini (gemini-3-flash-preview) [no-timeout]...
❌ gemini CLI not available: 'gemini' Popen failed: FileNotFoundError: [Errno 2]
   No such file or directory: 'gemini'
✅ Message sent to Qa-Engineer (ID: 777)   <- typed [Bridge Error] reply
✓ Message 776 acknowledged                 <- original consumed despite total failure
```

Three defects in the generic `ai_agent_bridge process <id>` path:

1. **Ignores the recipient.** The message was addressed to `cursor`; `process` unconditionally routes to Gemini. The per-agent `process-*` variants exist only for claude/codex/grok/grok-build/kimi — there is no recipient-derived routing and no cursor variant.
2. **Retired transport + stale slug.** It shells out to the standalone `gemini` CLI (retired — Gemini-family work routes through AGY/ACP) with hardcoded `gemini-3-flash-preview` — the same independently-hardcoded-default drift class as #6894.
3. **Consumes on failure.** After the CLI failed, it still sent an error reply AND acknowledged #776, so the queue looks drained while no analysis happened. A failed processing attempt must leave the message unconsumed (or mark it failed-needs-retry), never ack it.

Root-cause ask: derive the processing route from the message's `To:` seat via the same live routing/ACP layer the ask-* commands use (one source of truth, no per-command hardcoded model slugs), and make ack strictly conditional on a successful reply from the routed seat. Sibling sweep: audit the other `process-*` variants for the same ack-on-failure shape.

Context: QA-engineer intends to keep dropping advisor notes to seats without standing driver sessions ("nothing is draining" — their words, msg 776 item 3), so this path is about to become load-bearing.

Refs #6894 (stale-slug sibling), #5744 (context for the note being processed).

krisztiankoos:
Fixes #6915
Fixes #6894

## Root cause (one family, both issues)

Bridge legacy paths carried their own hardcoded routing/model state that drifted from the live ACP participant registry (`ACPX_SUPPORTED_PARTICIPANTS`, `scripts/agent_runtime/adapters/acpx.py:585`).

## Reproductions (unit level, before → after)

**#6915** — cursor-addressed advisor note from qa-engineer, transport fails:
- Before: routed to the retired gemini CLI with `gemini-3-flash-preview` regardless of `To: cursor`; sent `[Bridge Error]` reply AND `✓ Message acknowledged` — `acked=1` after total failure.
- After: `Routing message #N to ACP participant 'cursor' (recipient seat 'cursor', model=registry pin)...` → failure leaves `acked=0`, typed error reply still sent, status recorded `failed:...`. Success path: routed reply delivered to sender, then ack.

**#6894** — bare `ask-gemini`:
- Before: default `--model = 'gemini-3-flash-preview'` → `InterAgentTransportError: ACP participant 'agy' only supports its registered model pin 'gemini-3.7-flash-high'`.
- After: default is `None` → the route resolver applies the live registry pin (`route accepted with registry pin: 'gemini-3.7-flash-high'`); legacy `gemini*` slugs map to the current pin.

## Changes

- **New `scripts/ai_agent_bridge/_process.py`** — `process_message_for_recipient`: derives the route from the message's `To:` seat via `_acp_compat` (same layer the ask-* ACP routes use), defaults model to the registry pin, acks only after a successful routed reply, leaves failures unconsumed/retryable with the honest typed-error reply preserved. Already-acked messages are skipped.
- **`_cli.py`** — `process`, `process-all`, and interactive `process` rewired to the recipient-derived processor; `--model` defaults removed (`None` = registry pin); static `_LEGACY_GEMINI_TO_AGY_MODEL` table deleted; `ask-gemini` maps legacy slugs via `resolve_compat_model`; `ask-agy --to-model` help no longer hardcodes the pin.
- **`_acp_compat.py`** — new `registered_participant_model` / `resolve_compat_model`: defaults derive from the live registry; pin rotation is covered by test (simulated rotation moves every legacy default with it).
- **`_agy.py`** — `_DEFAULT_AGY_MODEL` hardcode replaced by `_default_agy_model()` reading the registry pin.
- **Ack-on-failure sweep (all process-* variants audited):** removed `acknowledge()` from every failure handler — `_gemini._send_gemini_error`, `_claude` (rate-limit/timeout/unavailable/review-checkout/generic/fallback), `_codex._handle_codex_error`, `_grok_build` error + incomplete-turn, `_kimi._handle_kimi_error`, `_agy._handle_agy_error`, `_opencode` incomplete-turn. All retain the typed error reply + `record_ask_failure`. `_cursor`/`_hermes` already raised without acking. Stuck rows remain reclaimable via `broker_cleanup --max-age`.

## Tests

- New `tests/ai_agent_bridge/test_process_routing.py` (21 tests): recipient-derived routing (cursor, gemini→agy), ack-only-after-reply ordering, failure/exception/empty-response/unroutable-recipient all leave the message unconsumed, already-acked skip, registry-pin default, simulated pin rotation, per-seat error-handler no-ack sweep.
- Mutation-checked: re-adding `acknowledge()` to a failure path turns 3+ of these tests red (verified for `_process._notify_processing_failure` and `_codex._handle_codex_error`).
- Updated `test_ab_discuss_bugs.py` (mapping now asserts live pin) and `test_coverage_misc.py` (`_send_gemini_error` must not ack).

## Verification

- `pytest tests/ai_agent_bridge/ tests/test_coverage_misc.py tests/test_kimi_integration.py tests/test_legacy_bridge_telemetry.py tests/agent_runtime/test_acpx_adapter.py` → **739 passed, 1 failed**; `tests/agent_runtime/ tests/test_dispatch_telemetry.py` → **394 passed, 2 failed**. All 3 failures reproduce on unmodified `main` in this worktree (`sealed_acp_python_missing` and 2 agy writer-prompt tests — environmental: dispatch worktrees have no `.venv`, which the worktree contract forbids creating).
- `ruff check` clean on all touched files.

## Known residuals (out of scope, not regressions)

- `converse` still uses the module-level legacy gemini path with a hardcoded `gemini-3.1-pro-preview` default — same drift class, but it is neither a `process-*` variant nor an `ask-*` seat.
- qa-engineer's reply surface (send-only identity) stays as-is: error replies to QA land in the broker while QA's monitored surface is its GitHub loci; deciding QA's reply surface explicitly is a policy call left to the operator (see #6915 comment).

No auto-merge armed; workers neither merge nor arm.

### DIFF
--- a/scripts/ai_agent_bridge/_acp_compat.py
+++ b/scripts/ai_agent_bridge/_acp_compat.py
@@ -44,6 +44,43 @@ def require_compat_target(command_target: str) -> str:
         ) from exc
 
 
+def registered_participant_model(participant: str) -> str | None:
+    """Return the live model pin for an ACP participant from the adapter registry.
+
+    The registry (``ACPX_SUPPORTED_PARTICIPANTS``) is the single source of
+    truth the route resolver enforces; consumers must read it instead of
+    carrying an independently hardcoded slug that goes stale on every model
+    rotation (#6894). Returns None for unknown or deliberately unpinned seats.
+    """
+    from agent_runtime.adapters.acpx import ACPX_SUPPORTED_PARTICIPANTS
+
+    entry = ACPX_SUPPORTED_PARTICIPANTS.get(participant)
+    if not entry:
+        return None
+    model = entry.get("model")
+    return str(model) if model else None
+
+
+def resolve_compat_model(command_target: str, model: str | None) -> str | None:
+    """Resolve a legacy ``--model`` value against the live participant registry.
+
+    ``None`` stays ``None`` so the route resolver applies the participant's
+    registered pin — a default can never drift from the registry (#6894).
+    Legacy ``gemini*`` slugs (including retired display labels) aimed at the
+    AGY seat map to that seat's current pin, the only model the seat accepts.
+    Anything else passes through unchanged and is validated loudly by the
+    route resolver.
+    """
+    participant = require_compat_target(command_target)
+    if not model:
+        return None
+    if participant == "agy":
+        pin = registered_participant_model("agy")
+        if pin and model != pin and model.strip().lower().startswith("gemini"):
+            return pin
+    return model
+
+
 # Per-seat default hard timeouts for compat asks (#6877). The generic 300s
 # ceiling is mis-sized for Kimi: K3 is a max-effort-only model whose long
 # deliberation before first output is designed behavior, and the fleet routes
--- a/scripts/ai_agent_bridge/_agy.py
+++ b/scripts/ai_agent_bridge/_agy.py
@@ -42,7 +42,20 @@ from ._review_worktree import (
 
 _DEFAULT_AGY_BRIDGE_TIMEOUT_SECONDS = 900
 _NO_TIMEOUT_AGY_BRIDGE_TIMEOUT_SECONDS = 24 * 60 * 60
-_DEFAULT_AGY_MODEL = "gemini-3.7-flash-high"
+
+
+def _default_agy_model() -> str:
+    """The AGY seat's model, read from the live ACP participant registry.
+
+    Never an independently hardcoded slug: the registry pin is what the route
+    resolver enforces, so deriving the default from it makes stale-slug drift
+    impossible (#6894).
+    """
+    from agent_runtime.adapters.acpx import ACPX_SUPPORTED_PARTICIPANTS
+
+    pin = ACPX_SUPPORTED_PARTICIPANTS["agy"]["model"]
+    assert pin is not None  # the agy participant is always pinned
+    return pin
 
 
 def _agy_ask_scratch_cwd() -> Path:
@@ -124,7 +137,7 @@ def ask_agy(
         # Refuse formal/exact-target review on AGY before any worktree or send.
         raise ValueError(AGY_SEALED_REVIEW_UNSUPPORTED)
     effective_model = resolve_model_selection(
-        lane="ask-agy", to_model=to_model, model=None, default=_DEFAULT_AGY_MODEL
+        lane="ask-agy", to_model=to_model, model=None, default=_default_agy_model()
     )
     msg_id = send_message(
         content,
@@ -188,7 +201,7 @@ def process_for_agy(
 
     _ = new_session  # No-op for now; Agy bridge calls are always fresh.
     timeout_val = _resolve_agy_bridge_timeout(no_timeout)
-    model = _extract_target_model(msg) or _DEFAULT_AGY_MODEL
+    model = _extract_target_model(msg) or _default_agy_model()
     effort = requested_effort(msg)
     effort_applied, effort_reason = unsupported_effort_note(
         lane="agy",
@@ -387,7 +400,11 @@ def _extract_target_model(msg: dict) -> str | None:
 
 
 def _handle_agy_error(msg: dict, message_id: int, reason: str) -> None:
-    """Record an Agy failure as a response message and acknowledge."""
+    """Record an Agy failure as a response message; no ack (#6915).
+
+    The inbound message stays unacknowledged so the failure is retryable;
+    acknowledgement happens only after a successful routed reply.
+    """
     print(f"\n❌ Agy error for message #{message_id}: {reason}")
     from ._ask_contract import failed_response_provenance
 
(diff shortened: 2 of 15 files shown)


krisztiankoos:
**Cross-family review of record — Gemini (gemini-3.7-flash-high via AGY, full diff): PASS.** Author kimi (Moonshot) → cross-family gate satisfied.

Reviewer verified the load-bearing semantics directly: tests assert consumption state in SQLite (`acked==0` on every failure shape, `==1` only on a valid routed reply), the success path is order-trapped (`ask` strictly before `ack`), every native error handler is parameterized as never-acking, and the pin-rotation test proves defaults track the registry without code edits. Happy-path ask-* routes verified regression-free. The author's declared out-of-scope residual (`converse` legacy default) is an acceptable boundary — banked below.

Review gate passed → arming auto-merge; merges on green blocking CI. Per #M-19: bridge is per-invocation CLI — no service bounce; live for callers on next pull.

krisztiankoos:
CI failures from run 32009066415 fixed in 86ebb7aad5 (pushed to this branch).

**Fix 1** — `tests/test_gemini_bridge.py::test_handle_ask_gemini_routes_to_agy` pinned the literal slug `gemini-3.6-flash-high`, exactly the drift class this PR eliminates. It now asserts against `registered_participant_model('agy')`, read from the same `ACPX_SUPPORTED_PARTICIPANTS` registry the route resolver enforces — no literal slug, so pin rotation can never stale it again.

**Fix 2** — `tests/test_codex_bridge.py::test_other_errors_still_ack_inbound` encoded the old ack-on-failure contract this PR deliberately removed. Inverted (renamed to `test_other_errors_leave_inbound_unconsumed`), not deleted: the parameterized never-ack coverage in `test_process_routing.py` calls `_handle_codex_error` directly, but nothing else covers the `AgentStalledError → _handle_codex_error` dispatch inside `process_for_codex` (the rate-limit sibling covers only `RateLimitedError`). It now asserts the new contract: inbound stays unconsumed (ack=0) while the error reply is still sent. No production-code changes.

**Evidence:** `pytest tests/test_gemini_bridge.py tests/test_codex_bridge.py tests/ai_agent_bridge/test_process_routing.py` → **49 passed in 5.48s**; `ruff check` on the three files → all checks passed.

**Why the local "739 passed in the bridge suites" run missed these:** the selection gap was directory shape, not test intent — the local run targeted the bridge package directory `tests/ai_agent_bridge/` (where the new `test_process_routing.py` lives) plus adjacent runtime suites, but the two failing files are legacy top-level flats — `tests/test_gemini_bridge.py` and `tests/test_codex_bridge.py` — which sit outside that glob and were never collected. The next bridge PR should run both levels: `pytest tests/ai_agent_bridge/ tests/test_gemini_bridge.py tests/test_codex_bridge.py` (or simply `pytest tests -k bridge`) so legacy flats and the package dir are covered in one selection.

krisztiankoos:
**Exact-head re-stamp — Gemini PASS at delta head `86ebb7a`**: both test-contract updates correctly encode the already-approved semantics (unconsumed-on-failure under AgentStalledError with error-reply assertions intact; registry-derived pin assertion), no production change smuggled in. Review gate holds at the exact head → arming auto-merge on green CI.

PR #6919, opened 2026-08-17
fix(ci): cheap-exit frontend when change denominator untouched

krisztiankoos:
## Summary
- Keep `frontend` / `frontend-e2e` **unconditionally scheduled** (no job-level `if:` skip) so CI Gate still sees `success`, never `skipped`.
- After checkout, a first-step path check (`scripts/ci/frontend_change_scope.py`) diffs against `pull_request.base.sha || merge_group.base_sha || github.event.before`; out-of-scope diffs log one loud decision line + `$GITHUB_STEP_SUMMARY` and exit 0; remaining steps are gated on `steps.scope.outputs.run`.
- Single-source denominator: `scripts/ci/frontend_change_denominator.json` (`site/`, `scripts/atlas/`, `scripts/lexicon/`, hydrate audit entrypoints, `.python-version`, `.github/workflows/ci.yml`).
- Drift guard pytest resolves hydrate entrypoints from `site/package.json` and asserts each resolved path stays inside that denominator. Pytest plan/shards and `ci-gate.needs` are unchanged.

Fixes #6917

## Changes
- `.github/workflows/ci.yml` — scope step + step-level gates on `frontend` / `frontend-e2e` (`fetch-depth: 0`)
- `scripts/ci/frontend_change_denominator.json` — shared denominator
- `scripts/ci/frontend_change_scope.py` — matcher, hydrate resolver, CI CLI
- `tests/test_frontend_change_scope.py` — unit + drift guard + workflow shape assertions

## Testing
- [x] `actionlint .github/workflows/ci.yml` — green
- [x] `ruff check scripts/ci/frontend_change_scope.py tests/test_frontend_change_scope.py` — clean
- [x] `pytest tests/test_frontend_change_scope.py tests/test_ci_queue_starvation.py tests/test_ci_changed_tests.py` — 25 passed
- [ ] CI green on this PR

### Mutation check (#M-4)
Remove `scripts/atlas/` from the denominator locally → guard fails; restore → guard passes:

```
=== MUTATION CHECK (remove scripts/atlas/ from denominator) ===
GUARD FAILED AS EXPECTED:
hydrate entrypoints outside frontend change denominator: scripts/atlas/atlas_db.py (denominator version=1)

=== RESTORE (full denominator) ===
[...]

PR #6931, opened 2026-08-17
fix(ci): use merge-base scope for frontend cheap-exit on PRs

### LINKED ISSUE #6930: [infra][ci] frontend cheap-exit never fires on PRs with a moved base — scope diff needs merge-base semantics
Live evidence from the first real script-only PR after #6919 merged (PR #6928, bridge-only changes): the frontend job ran the FULL 7m12s build (run 32010756237, job 95329593808 — the "Decide whether frontend work is in scope" step succeeded and then every build step ran).

Root cause, reproduced locally against the same refs:

```
merge-base(main, PR-head) = 405bc073
git diff --name-only <main-tip> <pr-head>   → 38 files   (two-dot: includes MAIN-side changes)
git diff --name-only <merge-base> <pr-head> → 15 files   (the PR's actual changes — zero denominator hits)
```

The main-side portion of the two-dot diff contains `site/package.json`, `site/package-lock.json`, `scripts/lexicon/runner/atlas_job.py` — denominator entries that were merged to main AFTER the PR branched. So for any PR whose base has advanced past a denominator-touching merge (at current cadence: almost all PRs), the scope check decides `run` and the cheap-exit never fires. Failure direction is SAFE (over-running, never under-running) but the optimization is silently defeated.

Fix at the helper layer (`scripts/ci/frontend_change_scope.py` + its ci.yml invocation):
- For `pull_request` / `merge_group`: diff must use **merge-base semantics** (`git merge-base BASE HEAD` then `diff MB..HEAD`, or three-dot `BASE...HEAD`) so only the PR's own changes count.
- For `push` (base = `github.event.before`): the current two-dot is correct (linear history) — keep it.
- Fail-safe unchanged: unresolvable merge-base (shallow clone!) → `decision=run`. Note the checkout step's fetch depth must be sufficient for merge-base to resolve — if the job uses a shallow checkout, deepen or fetch the base ref explicitly, and treat failure as run.
- Guard test: fixture repo where main advances with a denominator file while the branch touches only out-of-scope files → decision MUST be `cheap_exit`; plus the inverse (branch itself touches a denominator file after base moved → `run`).

Refs #6917, PR #6919, evidence run 32010756237.

krisztiankoos:
## Summary
- Frontend / frontend-e2e scope now uses **merge-base** range semantics for `pull_request` / `merge_group`, and keeps linear **two-dot** `before..after` for `push`.
- Scope step passes the **event head SHA** (`pull_request.head.sha` / `merge_group.head_sha`), not checkout `HEAD` (the PR merge commit). Diffing against the merge commit was the live failure mode on #6928 / run 32010756237.
- Unresolvable merge-base fails open to `decision=run`. Checkout already uses `fetch-depth: 0` (full history); comments document that shallowing this job would break merge-base unless the base ref is deepened/fetched.

## Evidence (PR #6928 refs)
- BEFORE (base…merge-commit): 30 files, denom hits `site/package.json`, `site/package-lock.json`, `scripts/lexicon/runner/atlas_job.py` → `decision=run`
- AFTER (`--event pull_request --head <pr.head.sha>`): 15 files, matched=0 → `decision=cheap_exit`
- Mutation: force two-dot on the moved-base fixture → falsely `decision=run` (guard test)

## Test plan
- [x] `/Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/python -m pytest tests/test_frontend_change_scope.py`
- [x] ruff check/format clean on owned paths
- [x] actionlint on `.github/workflows/ci.yml`
- [ ] CI green on this PR (frontend should cheap-exit if only helper/tests/workflow-scope touched — note: this PR **does** touch `.github/workflows/ci.yml`, which is in the denominator, so frontend **will run** here by design)

Fixes #6930

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -728,13 +728,23 @@ jobs:
           persist-credentials: false
           fetch-depth: 0
 
+      # fetch-depth: 0 (full history) so merge-base(base, head) resolves for
+      # PR/merge_group scope (#6930). Do not shallow this job without also
+      # deepening/fetching the base ref in the scope step.
       - name: Decide whether frontend work is in scope
         id: scope
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.merge_group.base_sha || github.event.before }}
+          # Event head — not checkout HEAD. PR checkouts are the merge commit;
+          # three-dot/merge-base against that invents main-side denominator hits.
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.event.merge_group.head_sha || github.sha }}
+          EVENT_NAME: ${{ github.event_name }}
         run: |
           set -euo pipefail
-          python3 scripts/ci/frontend_change_scope.py --base "${BASE_SHA:-}"
+          python3 scripts/ci/frontend_change_scope.py \
+            --base "${BASE_SHA:-}" \
+            --head "${HEAD_SHA:-HEAD}" \
+            --event "${EVENT_NAME:-}"
 
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         if: steps.scope.outputs.run == 'true'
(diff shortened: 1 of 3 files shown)


krisztiankoos:
**Dual-family quorum COMPLETE (cursor:auto author): Gemini PASS + Grok PASS**, both against the full inline diff at head.

The load-bearing question (can branch-only scoping miss a merged-combination frontend bug?) got a crisp answer from the grok seat, preserved for the record: main-side denominator hits were already frontend-tested when they merged, so skipping on them is correct for a non-gating runner filter; combination-only bugs are a denominator/always-on-contracts concern, not a reason to re-count main-side files — and every unresolvable-range path fails open to `run`. Honest caveat noted: merge_group cheap-exits identically (it is not a hidden safety net), accepted on the same grounds. Fixture guards verified mutation-sensitive; all #6919 properties preserved; `--event` now required by the workflow-shape test so a bare local invocation over-runs rather than under-runs.

Review gate passed → arming auto-merge; merges on green blocking CI (frontend runs FULL on this PR by design — it touches ci.yml, which is in the denominator). The real cheap-exit proof lands on the next bridge/scripts-only PR.

krisztiankoos:
## CI fix: #5740 subprocess timeout guard

Root cause: fixture helper `_git()` in `tests/test_frontend_change_scope.py` called `subprocess.check_output` without `timeout=`.

Fix in `06e1247ebf`: added `timeout=30` to that helper (only change).

### Acceptance
```
/Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/python -m pytest tests/test_frontend_change_scope.py tests/test_subprocess_timeout_guard.py -q --tb=short
→ 23 passed in 6.42s

ruff check + ruff format --check tests/test_frontend_change_scope.py
→ clean
```

Pushed to `cursor/infra-6930-mergebase-scope`. No auto-merge armed.

krisztiankoos:
Driver note: the `06e1247` delta adds only `timeout=30` to the test fixture's git helper (the #5740 guard requirement) — trivial, test-only, no semantic change; accepted under the standing dual-family quorum without a re-stamp. Auto-merge remains armed.

PR #6932, opened 2026-08-17
fix(bridge): resolve converse gemini default via the live registry pin

### LINKED ISSUE #6929: [infra][bridge] converse command still carries a hardcoded legacy gemini default — align with the registry-derived resolution from #6928
Author-declared residual from PR #6928 (which fixed #6915/#6894): the `converse` command's module-level legacy path still hardcodes a gemini default model instead of resolving via `resolve_compat_model`/the live ACP participant registry. Same drift class as #6894 — one small PR to route it through the #6928 resolver. Refs #6928 #6894.

krisztiankoos:
Fixes #6929

## Root cause

Author-declared residual from PR #6928: the `converse` command still hardcoded `gemini-3.1-pro-preview` on both the CLI `--model` default and `converse_gemini()`'s signature. Same drift class as #6894 — an independently hardcoded slug that goes stale on every AGY pin rotation.

## Change

No new mechanism. Alignment onto the #6928 resolver:

- `converse --model` defaults to `None` (never an independently hardcoded slug).
- The handler runs `resolve_compat_model("gemini", model)` so legacy `gemini*` slugs remap to the live AGY pin.
- `converse_gemini()` applies that same resolver, then `registered_participant_model("agy")` when the caller omitted `--model`.

Converse path slug sweep: the only other hardcoded model strings in this path were those two defaults. Nearby `--from default="gemini"` is a seat name, not a model slug. `ask_gemini` / `process_and_respond` still default to `GEMINI_DEFAULT_MODEL` (FLASH / `gemini-2.0-flash`) — out of converse scope, not this PR.

## Tests

- Bare `converse` parser default is `None`; `converse_gemini` signature default is `None` (mutation-checked).
- Simulated pin rotation (`gemini-9.9-flash-high`) moves the converse default and remaps `gemini-3.1-pro-preview`; non-legacy slugs pass through.
- Coverage lives in `tests/ai_agent_bridge/test_process_routing.py` and the top-level flat `tests/test_gemini_bridge.py` that #6928's selection missed.

Mutation-check: restoring the hardcoded default turns all four new tests red.

## Verification

- `ruff check` clean on all touched files.
- Targeted: `tests/test_gemini_bridge.py` + `tests/test_codex_bridge.py` + `tests/ai_agent_bridge/test_process_routing.py` → **60 passed**.
- Broader `tests/ai_agent_bridge/` → **447 passed**, 1 failed: `test_sealed_acp_config_is_parent_owned_and_snapshot_pinned` (`sealed_acp_python_missing` looking for a worktree `.venv`). Same environmental failure #6928 recorded on unmodified main; worktree contract forbids creating `.venv` here.

No auto-merge armed; workers neither merge nor arm.

### DIFF
--- a/scripts/ai_agent_bridge/_cli.py
+++ b/scripts/ai_agent_bridge/_cli.py
@@ -965,7 +965,12 @@ def _build_parser() -> argparse.ArgumentParser:
     converse_parser = subparsers.add_parser("converse", help="Multi-turn conversation with Gemini (includes history)")
     converse_parser.add_argument("content", help="Message content (use '-' to read from stdin)")
     converse_parser.add_argument("--task-id", required=True, help="Conversation thread ID (e.g., 'a1-1-planning')")
-    converse_parser.add_argument("--model", default="gemini-3.1-pro-preview", help="Gemini model")
+    converse_parser.add_argument(
+        "--model",
+        default=None,
+        help="Legacy Gemini model slug; mapped to the AGY seat's live registry pin "
+        "(default: the pin itself — never an independently hardcoded slug)",
+    )
     converse_parser.add_argument(
         "--no-github", dest="no_github", action="store_true", help="Skip auto-posting to GitHub"
     )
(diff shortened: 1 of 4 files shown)


krisztiankoos:

Confirmed: both hardcoded slugs replaced through the merged #6928 resolver with no new mechanism, explicit --model behavior unchanged, the four new tests pin the registry-derived defaults via dynamic pin-rotation (regression to any literal fails the suite), and the diff-context sweep found no remaining hardcoded model slugs on the converse path. The out-of-scope sibling (GEMINI_DEFAULT_MODEL internals) stays banked on #6929.

Review gate passed → arming auto-merge; merges on green blocking CI.

Issue #6921, opened 2026-08-17
[infra][runtime] watchdog activity-poller follow-ups from the #6897 CF review: child-subprocess CPU weighting + sustained-over-N-polls smoothing


krisztiankoos:
Banked from the gemini CF review of PR #6897 (merged): two non-blocking hardening directions for the process-activity poller now that the 0.5s/5s trip bar landed. (1) Weight child-subprocess CPU separately from the root CLI process — a wedged root with a busy unrelated child should not count as activity. (2) Consider sustained-activity-over-N-polls smoothing instead of a single-window threshold, so one legitimate 0.6s blip in an otherwise dead session cannot indefinitely reset silence detection (inverse of the #6888 failure). Also still on watch from the #6888 author's honest residual: a wedged CLI that heartbeat-writes its --log-file would defeat the mtime poller — if an agy stall recurs POST-#6897, check the log-file mtime channel first. Low priority; evidence thresholds must be re-measured per #6897's methodology before changing constants. Refs #6897 #6888.

krisztiankoos:
## Measurement verdict (grok lane, read-only campaign 2026-08-17 — full note below; driver-posted)

## 1. What #6897 set and how it measured

Merged PR #6897 (`1c5b19277dd9e3eb8dedfbcff0d3592832d06f05`, 2026-08-16) raised `_PROCESS_ACTIVITY_MIN_CPU_SECONDS` from `0.05` to `0.5` per 5s poll.

Quoted methodology (PR body + `watchdog.py` comment): live idle `agy` pid **52087**, `psutil.Process.cpu_times()` user+system, 5s windows. Result then: **steady ~0.03s/5s**, **spike 0.0657s in a 30s sample**. That spike beat the old 0.05s bar and reset `last_activity` forever.

The poller is **per-process, not tree-sum**. Any one of `root + children(recursive=True)` with `Δcpu >= 0.5` in a 5s window refreshes `last_activity`. Disk I/O is a second channel: any `io_counters` tuple change counts as activity (`watchdog.py` `_process_tree_has_activity`).

#6921 banks two Gemini CF follow-ups: (i) weight children separately so a wedged root + busy unrelated child does not count; (ii) require N consecutive over-bar polls so one 0.6s blip cannot reset a 3600s silence clock. Also the #6888 residual: a wedged CLI that heartbeat-writes `--log-file` would still defeat the mtime poller.

---

## 2. Machine and method (this re-measure)

```text
[...]

PR #6934, opened 2026-08-17
fix(runtime): scope grok liveness off shared ~/.grok mtime (#6933)

### LINKED ISSUE #6933: [infra][runtime] grok liveness signal is cross-session contaminated — shared ~/.grok dir mtime bumped by OTHER grok sessions
Found during the #6921 read-only measurement campaign (2026-08-17, grok lane): while a supervised grok session's own event log (atlas events.jsonl) stayed frozen for the whole 36×5s campaign, the shared ~/.grok liveness directory's mtime was bumped TWICE by a different concurrently running grok session (the measuring task itself). Consequence: any watchdog/liveness check that reads the shared directory (rather than the supervised session's own artifacts) can be kept alive by an unrelated grok session — masking a wedged grok exactly the way #6888's log-heartbeat residual anticipated, but cross-session. Ask: identify what (if anything) in the watchdog/monitoring stack consumes ~/.grok mtimes for liveness; if anything does, scope it per-session (session-specific file/dir), and add a guard test with two simulated sessions. If nothing consumes it, close with that evidence. Refs #6921 #6897 #6888. Evidence: /tmp/watchdog-6921-measure.json (raw campaign, uncommitted).

krisztiankoos:
## Summary
- **Verdict:** the runtime watchdog *does* consume shared `~/.grok` mtimes for liveness — `GrokBuildAdapter._liveness_paths()` returned `(Path.home() / ".grok",)`.
- Scope the signal to the cwd-keyed sessions parent plus this invocation’s own session dir / `events.jsonl` (resume id, or a session created after the build-time snapshot). Never watch `GROK_HOME` / `logs/unified.jsonl`.
- Add a two-session guard test; mutation-checked: restoring the shared-home path fails `assert grok_home not in paths`.

## Trace (consumers)
| Path | Role |
| --- | --- |
| `scripts/agent_runtime/adapters/grok_build.py` `_liveness_paths` → `runner.start_watchdog` | **was** the contaminated consumer; fixed |
| `scripts/agent_runtime/watchdog.py` `_mtime_poller` | generic poller; only sees adapter-provided paths |
| Hygiene / inventory / hook-profile `~/.grok` readers | disk size / session inventory / config — **not** liveness |
| `GROK_HOME` in layerb / bridge | scoped homes for isolation / events lookup — **not** watchdog liveness |

Evidence from `/tmp/watchdog-6921-measure.json`: supervised atlas `events.jsonl` frozen (0 bumps) while shared `~/.grok` bumped twice and a peer session’s `events.jsonl` bumped — exactly the false-alive channel.

## Test plan
- [x] `pytest tests/test_grok_build_adapter.py` (33 passed)
- [x] Mutation: shared-home revert fails the new guard assertion

Fixes #6933

### DIFF
--- a/scripts/agent_runtime/adapters/grok_build.py
+++ b/scripts/agent_runtime/adapters/grok_build.py
@@ -124,6 +124,27 @@ def validate_grok_effort(effort: str | None) -> str | None:
     return effort
 
 
+def resolve_grok_home(*, env: dict[str, str] | None = None) -> Path:
+    """Return the active Grok home (``GROK_HOME`` or ``~/.grok``)."""
+    if env is not None and env.get("GROK_HOME"):
+        return Path(env["GROK_HOME"])
+    configured = os.environ.get("GROK_HOME")
+    if configured:
+        return Path(configured)
+    return Path.home() / ".grok"
+
+
+def grok_cwd_sessions_dir(grok_home: Path, cwd: Path) -> Path:
+    """Return the cwd-keyed sessions parent under ``grok_home``.
+
+    Native Grok stores one directory per session beneath
+    ``sessions/<urlquoted-resolved-cwd>/``. Peer sessions that share a cwd
+    still get distinct child directories; the shared ``GROK_HOME`` root must
+    never be treated as a liveness signal (#6933).
+    """
+    return grok_home / "sessions" / quote(str(cwd.resolve()), safe="")
+
+
 def grok_session_dir(grok_home: Path, cwd: Path, session_id: str) -> Path:
     """Return Grok's session directory for ``cwd`` and ``session_id``.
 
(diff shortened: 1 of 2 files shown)


krisztiankoos:
**Dual-family quorum COMPLETE (cursor:auto author): Gemini PASS + Grok PASS**, both on the full inline diff at head.

The grok seat judged it against real grok session behavior: shared-home liveness is gone, resume binding and pre-existing same-cwd peers are handled correctly, quoting matches the native layout. Its non-blocking findings — the newest-post-snapshot re-pick can bind a LATER same-cwd peer (narrow residual race), instance-only snapshot, pre-mkdir startup window — are banked as #6935 (pin the discovered session id onto plan.metadata) rather than a respin: the shipped fix removes the vector that actually bit in #6921's measurements.

Review gate passed → arming auto-merge; merges on green blocking CI. This PR is scripts-only — it should also be the live proof of the #6931 merge-base cheap-exit (expect `decision=cheap_exit` and a seconds-long frontend job).

PR #6936, opened 2026-08-17
fix(runtime): pin grok liveness session onto plan.metadata (#6935)

### LINKED ISSUE #6935: [infra][runtime] pin the grok liveness session once discovered — newest-post-snapshot re-pick can bind a later same-cwd peer
Banked from the grok quorum seat on PR #6934 (PASS, non-blocking): the session-selection heuristic re-picks the newest post-snapshot session each poll, so a same-cwd grok session started AFTER the supervised build can steal the liveness bind — a narrower recurrence of #6933's contamination (rare: needs same-cwd concurrent starts, which lane staggering mostly avoids). Prescribed hardening: pin the first-discovered session id onto plan.metadata once bound (also fixes the instance-only snapshot — a plan-only poller can then reproduce the bind), tolerate/document the pre-mkdir startup window against the stall timer, and add tests: newest-steals-bind, split-instance non-resume, GROK_HOME override. Refs #6934 #6933.

krisztiankoos:
## Summary
- Pin the first post-snapshot Grok session id onto `plan.metadata` once discovered, and stamp the build-time child-name snapshot on the plan, so later same-cwd peers cannot steal the liveness bind and a plan-only poller can reproduce it.
- Document the accepted pre-mkdir startup window against the stall timer (stale `sessions_root` until this child appears; missing paths already baseline at `0.0`).
- Add mutation-checked guards: newest-steals-bind, split-instance non-resume, and `GROK_HOME` / `env_overrides`.

## Changes
- `scripts/agent_runtime/adapters/grok_build.py`: metadata keys `liveness_session_id` + `liveness_session_dir_snapshot`; bound-id path after pin/resume
- `tests/test_grok_build_adapter.py`: three new #6935 guards + pin assertion on the peer/home test

## Testing
- [x] `/Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/python -m pytest tests/test_grok_build_adapter.py` — 36 passed
- [x] `ruff check` on touched files — clean
- [x] Mutation checks: clearing the pin lets a later sibling steal; dropping the plan snapshot lets a fresh adapter bind a preexisting peer; ignoring `GROK_HOME` resolves under `~/.grok`

## Related Issues
Fixes #6935
Refs #6934 #6933

### DIFF
--- a/scripts/agent_runtime/adapters/grok_build.py
+++ b/scripts/agent_runtime/adapters/grok_build.py
@@ -348,7 +348,6 @@ class GrokBuildAdapter:
 
         snapshot = self._reset_per_invocation_state(
             cwd=execution_cwd,
-            resume_session_id=resume_session_id,
             env_overrides={},
         )
         metadata: dict[str, object] = {
@@ -457,19 +456,21 @@ class GrokBuildAdapter:
         )
         if discovered_id and not bound:
             # Mutate the plan-owned dict (InvocationPlan is frozen, metadata
-            # contents are not) so subsequent polls keep this bind.
+            # contents are not) so subsequent polls keep this bind. Never pin
+            # onto adapter instance state — a shared adapter serving two plans
+            # would otherwise hand plan B the session id pinned for plan A.
             metadata[_META_LIVENESS_SESSION_ID] = discovered_id
-            self._liveness_session_id = discovered_id
         return paths
 
     def _bound_liveness_session_id(self, metadata: dict[str, object]) -> str | None:
-        """Return resume or pinned liveness session id, if any."""
+        """Return resume or pinned liveness session id from plan metadata only.
+
+        Instance-level bind is intentionally absent: a shared adapter can
+        poll multiple plans, and an instance pin contaminates unpinned peers
+        (#6935 FAIL delta).
+        """
         for key in (_META_RESUME_SESSION_ID, _META_LIVENESS_SESSION_ID):
             raw = metadata.get(key)
-            if isinstance(raw, str) and raw:
-                return raw
-        for attr in ("_resume_session_id", "_liveness_session_id"):
-            raw = getattr(self, attr, None)
             if isinstance(raw, str) and raw:
                 return raw
         return None
@@ -494,12 +495,9 @@ class GrokBuildAdapter:
         self,
         *,
         cwd: Path,
-        resume_session_id: str | None,
         env_overrides: dict[str, str],
     ) -> set[Path]:
         """Snapshot pre-existing cwd-scoped session dirs before launch."""
-        self._resume_session_id = resume_session_id
-        self._liveness_session_id = resume_session_id
         self._session_dir_snapshot = self._snapshot_preexisting_session_dirs(
             cwd, env_overrides=env_overrides
         )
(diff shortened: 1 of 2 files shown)


krisztiankoos:
**Quorum split resolved as FAIL-until-delta: Grok FAIL (domain seat, prescription author) overrides Gemini PASS.**

Blocking finding: the instance-level `_liveness_session_id` fallback bind reintroduces contamination one layer up — a SHARED adapter serving two plans hands an unpinned plan the session id pinned for another, silently keeping a later plan "live" off a peer (false-LIVE, worse than the false-dead deleted-dir case). Pin-on-metadata + snapshot-on-metadata are correct; the instance bind undoes them for the shared-adapter poller.

Required delta (dispatched to the author lane): `_bound_liveness_session_id` reads only `plan.metadata`; instance snapshot fallback may stay, instance bind goes; same-adapter two-plan guard (A pinned, B unpinned → B never gets A's id), mutation-checked. Re-review after.

krisztiankoos:
## FAIL delta applied (author lane)

Dropped instance-level `_liveness_session_id` / `_resume_session_id` fallback from `_bound_liveness_session_id`. Bind now reads **only** `plan.metadata`. Instance snapshot fallback for `liveness_session_dir_snapshot` remains. Discover no longer writes the pin onto adapter instance state.

### Mutation evidence
Restoring the instance bind read **and** `self._liveness_session_id = discovered_id` on discover makes `test_liveness_shared_adapter_two_plans_no_cross_pin` fail: after plan A pins, unpinned plan B on the same adapter resolves to A's session paths instead of `sessions_root` (false-LIVE).

With the fix: A pinned + B unpinned (A in B's snapshot) → B stays on `sessions_root`, then binds its own post-snapshot child — never A's id.

Also added optional `test_liveness_pinned_dir_deleted_mid_run_stays_bound` (false-dead on deleted pin is acceptable).

### Acceptance
- `pytest tests/test_grok_build_adapter.py` → 38 passed
- `ruff check` on touched files → clean
- head: `a6c37970fd` on `cursor/infra-6935-liveness-pin`

Ready for re-review. Not arming auto-merge.

krisztiankoos:
**Re-review at delta head `a6c3797`: Grok PASS (FAIL author satisfied) — with driver evidence-anchoring** (the seat's reply was substantive but tripped the non-evidentiary formality gate, so the concrete references are supplied here from a deterministic diff check):

- Both `self._liveness_session_id` assignment sites removed (discover-write and resume-write); the only remaining textual occurrence is the mutation-check docstring in `tests/test_grok_build_adapter.py`. `_reset_per_invocation_state` no longer carries `_liveness_session_id`/`_resume_session_id`.
- `_bound_liveness_session_id(metadata)` reads plan metadata only (resume, then `liveness_session_id`).
- The two-plan guard builds both plans first, then interleaves polls: asserts `"liveness_session_id" not in plan_b.metadata` while A is bound, B later pins its own `plan-b-session`, and A's pin stays `plan-a-session` throughout — exactly the isolation the FAIL demanded.
- Deleted-pin test documents the accepted false-dead direction.

Grok's non-blocking residual (instance `_session_dir_snapshot` fallback still shared) is accepted as part of the minimum fix — the bind path is isolated; snapshot only gates "new session" detection. Gemini's PASS from round 1 covered the pin lifecycle; quorum complete at this head → arming auto-merge on green CI.

PR #6940, opened 2026-08-17
fix(delegate): ignore concurrent sibling sandboxes in read-only guard

### LINKED ISSUE #6938: [infra][delegate] read-only mutation guard false-fails on concurrent sibling dispatch worktrees — twice on 2026-08-17
Two read-only dispatches today were marked failed with 'read-only checkout mutation detected: .worktrees/dispatch/<other-lane-task>/' while their deliverables completed intact: analyze-776-cursor (tripped on cursor/codeql-path-injection-fix appearing mid-run) and infra-6830-kill-forensics (tripped on cursor/infra-6106-caller-telemetry). The guard's mutation denominator evidently includes sibling dispatch-worktree creation under .worktrees/, so ANY concurrent dispatch falsifies a read-only task's status — statuses lie, drivers must read result files (burn class already documented). Fix at the guard layer: exclude .worktrees/** (other tasks' sandboxes) from a read-only task's mutation scan, or scope the scan to paths the task could actually touch. Guard test: concurrent worktree add during a read-only task ⇒ status stays done. Refs the two task records above.

krisztiankoos:
## Summary
- Read-only mutation guard no longer false-fails when a concurrent sibling `git worktree add` appears under `.worktrees/dispatch/<agent>/<task>/` (#6938).
- Scoping choice: exempt only **newly appeared** layout-A sandbox roots (absent from the pre-snapshot), not a blanket `.worktrees/**` exclude — so a write into an **already-visible** sibling sandbox still trips the guard when the root checkout can see it.
- Residual: writes into a sandbox created in the same window as the concurrent add are indistinguishable from sandbox creation and are not attributed; tasks whose cwd is already inside their own worktree never see sibling paths in the snapshot anyway.

## Changes
- `scripts/delegate.py`: detect `.worktrees/dispatch/<agent>/<task>` roots; skip mutations under roots that were absent pre-run.
- `tests/test_delegate.py`: unit mutation-check + concurrent `git worktree add` stays `done` + inverse task file write still `failed`.

## Testing
- [x] `/Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/python -m pytest tests/test_delegate.py` — 266 passed
- [x] `ruff check scripts/delegate.py tests/test_delegate.py` — clean
- [ ] Website builds without errors (`cd site && npm run build`) — N/A (delegate-only)
- [ ] Ukrainian text reviewed for naturalness — N/A

## Related Issues
Fixes #6938

### DIFF
--- a/scripts/delegate.py
+++ b/scripts/delegate.py
@@ -2279,6 +2279,11 @@ _READ_ONLY_RUNTIME_STATE_SUFFIXES = (
     ".sqlite3-wal",
 )
 _READ_ONLY_UNTRACKED_OR_IGNORED_STATUSES = frozenset({"??", "!!"})
+# Dispatch sandboxes live at ``.worktrees/dispatch/<agent>/<task>/`` (layout A).
+# Concurrent ``git worktree add`` under that prefix must not false-fail a
+# read-only task that is scanning the shared primary checkout (#6938).
+_READ_ONLY_DISPATCH_SANDBOX_PREFIX = (".worktrees", "dispatch")
+_READ_ONLY_DISPATCH_SANDBOX_ROOT_PARTS = 4  # .worktrees / dispatch / agent / task
 
 
 def _normalize_read_only_relpath(path: str) -> str:
@@ -2286,7 +2291,7 @@ def _normalize_read_only_relpath(path: str) -> str:
     normalized = path.replace("\\", "/")
     while normalized.startswith("./"):
         normalized = normalized[2:]
-    return normalized
+    return normalized.rstrip("/")
 
 
 def _is_read_only_runtime_telemetry_path(path: str) -> bool:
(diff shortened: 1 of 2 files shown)


krisztiankoos:
**Dual-family quorum COMPLETE (cursor:auto author): Gemini PASS + Grok PASS**, both against the full inline diff at head.

Both seats confirmed: the exemption is bounded to newly-appeared 4-part layout-A sandbox roots against a snapshot-pure comparison; the same-window residual is honestly named and prefix-bounded; the two inverse guards (task's own write, write under existing sibling) cover the mute/false-pass directions that matter. Grok's non-blocking notes banked for the record: the exemption includes the task's OWN newly-created sandbox (don't later describe it as "only other agents'"), the mute depends on snapshot completeness (don't weaken the snapshot's ignored-file recursion without revisiting), and the global trailing-slash strip is a small behavior widening outside #6938's strict scope — acceptable.

Review gate passed → arming auto-merge; merges on green blocking CI. This also un-lies the statuses of the two false-failed tasks going forward.

PR #6946, opened 2026-08-17
fix(atlas): per-host ATLAS_RUN_ROOT default for hramatka jobs

krisztiankoos:
## Summary
Remote/local class-B launchers default `ATLAS_RUN_ROOT` to `/home/ops/atlas-jobs` on `hramatka`/`vps`, otherwise `/home/ops/atlas-runner`. The remote wrapper `mkdir -p`s that root before scp. Nothing writes under `/opt/hramatka`.

## Test plan
- [x] `tests/test_launch_reenrich_target_detection.py` added/extended in the same commit
- [ ] CI green

Refs #6876

### DIFF
--- a/scripts/lexicon/runner/launch_reenrich_class_b.sh
+++ b/scripts/lexicon/runner/launch_reenrich_class_b.sh
@@ -2,7 +2,7 @@
 # Detached, idempotent launcher for #6369 Class-B residual EN re-enrich.
 # Mirrors launch_enrich.sh under MemoryHigh=1.5G MemoryMax=2.0G.
 #
-# Runs entirely against the atlas-runner *work-dir* — it never mutates the
+# Runs entirely against the remote-host *work-dir* — it never mutates the
 # VPS repo checkout at $REPO. This matters because the checkout there is
 # routinely stale/dirty (large data/ dirs deliberately deleted for disk
 # headroom); this launcher only ever reads from it (sources.db, kaikki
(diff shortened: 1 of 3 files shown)


krisztiankoos:
**Review: PR #6946 — per-host `ATLAS_RUN_ROOT` default (refs #6876)**

Scope checked: both class-B launchers, the remote `mkdir -p` ordering, and the new test coverage. All CI green (head `44772a2`).

**Correct:**
- `launch_reenrich_class_b_remote.sh` computes `RUN_ROOT` from `HOST` (exact match `hramatka|vps`), then `ssh_q`s `mkdir -p "$RUN_ROOT"` right after the reachability check (`ssh_q "true"`) and before the first outbound transfer (sources.db rsync, driver scp) — matches the stated intent for a fresh hramatka/vps host with no `atlas-jobs` dir yet.
- The remote wrapper forwards `ATLAS_RUN_ROOT=$RUN_ROOT` explicitly in `remote_cmd` (line 296) when invoking the on-host launcher, so the two scripts' *default* logic never actually races against each other in the normal call path — the remote wrapper's computed value always wins on-host via env override.
- No `/opt/hramatka` writes anywhere in the diff; test asserts this directly.
- `ATLAS_RUN_ROOT=""` (explicit empty) falls through to the per-host default in both scripts (`:-` semantics), consistent between the two.
- Quoting (`printf '%q'`) on the new `mkdir -p` matches the file's existing ssh-remote-command pattern.

**Minor inconsistency (non-blocking) — host matching is asymmetric between the two scripts:**
- `launch_reenrich_class_b.sh` (on-host hostname fallback) uses **glob** patterns: `hramatka*|vps*` — matches any hostname *starting with* `hramatka` or `vps` (e.g. a hypothetical `hramatka-2` or `vpsbackup` box).
- `launch_reenrich_class_b_remote.sh` (Mac-side, keyed on `$ATLAS_RUNNER_HOST`) uses **exact** match: `hramatka|vps`.
- The test suite encodes this asymmetry rather than flagging it: `test_remote_wrapper_run_root_defaults_per_host` expects `hramatka-2` → `atlas-runner` (exact-match fallthrough), while `test_local_launcher_run_root_defaults_by_hostname` expects `hramatka.example.com` → `atlas-jobs` (glob match).
- Given the remote wrapper always forwards an explicit `ATLAS_RUN_ROOT`, this only bites on **direct on-host invocation** of `launch_reenrich_class_b.sh` (bypassing the wrapper) on a host whose short name merely starts with `hramatka`/`vps` without being one of the two intended aliases — a narrow but real drift between "what the fleet considers hramatka/vps" in the two launchers. Suggest tightening the local launcher's case arms to exact match (`hramatka|vps`) for consistency, or if the wildcard is intentional (e.g. to cover `hramatka.local`/FQDN forms `hostname -s` might still return), say so in the comment — right now the two scripts silently disagree on what counts as "hramatka".

Not a blocker for this PR's stated scope (#6876 default + mkdir ordering), which is otherwise correctly implemented and tested.

**VERDICT: COMMENT**

Issue #6944, opened 2026-08-17
[4387][atlas] Monitor /api/atlas-jobs: host load + results + contracts (batched)


krisztiankoos:
Parent: #4387. Related (do **not** close): #6876.

Operator (2026-08-17): add easy/lightweight API support for VPS **load** and **results**; design with an advisor then implement; **check queued API changes and batch them together**.

## Inventory (open queue, 2026-08-17)

No open PR already implements VPS load/results. Live Monitor already has `/api/atlas-jobs` list / health / submit / GET `{id}` / close (PR #6867). That family is **missing** from `scripts/api/route_contracts.py` and `docs/MONITOR-API.md` (live `GET /api/contracts/routes` has no `atlas-jobs` key).

### In this batch (one design, one implement PR)

1. **New:** host load + results list on the existing Monitor facade (not a new daemon / bus / `/api/vps` unless advisors prove the existing prefix cannot carry it).
2. **Contract + docs gap from #6867:** register `/api/atlas-jobs` in `route_contracts.py` + `docs/MONITOR-API.md` + tests so the live router is no longer an undeclared family.
3. **Dual-host:** operator GO 2026-08-17 allows jobs on both `atlas-runner` and `hramatka`. Load must cover both named hosts. Dual-host protocol change is already in flight (`6876-claude-dual-host`); this API must compose with it, not fight it. GET `{id}` today defaults `host=atlas-runner`.

### Out of this batch (do not fold in)

| Issue | Why separate |
| --- | --- |
| #6941 | LaunchAgent Program wrapper (ops / LWCR). Same process as #6937, not a product API contract. |
| #6892 | grok-bot QA of the shipped Work board. Not an API change. |
| #6849 #6850 #6851 #6859 #6862 #6880 | Work API `/api/work/v1` — already merged/closed. |
| #5092 | VESUM inspection API — lexicon plane. |
| #4384 #4920 | Practice Hub product backend (accounts / FSRS). Different plane. |
| #6106 | Broker caller migration. |
| #6870 | grok-4.6 routing residuals. |

## User-visible outcome

An operator or driver can, from local Monitor (`:8765`), see **both named hosts' load** and a **results list** without SSH and without parsing journal files by hand. Responses never include raw IPs, credentials, or private host addressing.

## Non-goals

- New message bus, new daemon, or a second HTTP server.
- Closing #6876 (residual EN still open).
[...]

PR #6947, opened 2026-08-17
fix(atlas): allow reenrich jobs on hramatka alongside atlas-runner

krisztiankoos:
## Summary

Operator GO 2026-08-17: atlas jobs may run on both `atlas-runner` and
`hramatka`, not `atlas-runner` only.

- `ALLOWED_HOSTS['reenrich']` now includes `atlas-runner`, `hramatka`, and
  the `vps` ssh alias (confirmed same machine via
  `scripts/config/trails/estate.v1.yaml`: `vps` is `pilot-vps`, which also
  hosts `hramatka-api`).
- Dropped the `validate_plan` reject that refused batch kinds on
  hramatka/vps. Unknown host strings are still rejected (new test).
- Per-host default work root: `atlas-runner` keeps `/home/ops/atlas-runner`;
  `hramatka`/`vps` default to a new `/home/ops/atlas-jobs` (does not exist
  yet — submit/launcher `mkdir -p` it on first use). Kept separate from
  hramatka's teacher-facing `/opt/hramatka` and `/srv` trees — nothing here
  writes there. `require_safe_workdir` / `work_dir_for` / `_run_root` now
  take an optional `host` and forward it consistently; `submit()` also
  forwards `ATLAS_RUN_ROOT` to the remote launcher's env so
  `REMOTE_REPO` derivation matches the host actually targeted (complements
  #6946, which teaches the bash launchers the same per-host default).
- Fixed a latent bug in `SshHostAdapter.free_disk_bytes`: it ignored `host`
  when deriving its default probe path, which would have silently probed
  atlas-runner's root even for a hramatka disk-floor check.
- `status`/`pull` no longer refuse `hramatka`/`vps`.
- MemoryHigh/MemoryMax stay at the existing conservative 1.5G/2G on the
  class-B launcher (host-agnostic already) — documented in the runbook why
  that's load-bearing on hramatka (protects the teacher API/GH runners
  sharing that box).

## Test plan
- `pytest tests/test_atlas_job_protocol.py tests/api/test_atlas_jobs_router.py -q` — 54 passed
- `ruff check` on all changed files — clean

Refs #6876.

PR #6948, opened 2026-08-17
feat(api): atlas-jobs host load + results list

krisztiankoos:
Refs #6944

X-Agent: agy/6944-atlas-jobs-load


krisztiankoos:
## Cross-family review — PR #6948 (Claude, outside author family)

**Verdict: REQUEST_CHANGES**

### Freeze constraints — verified, all satisfied
- `GET /load` and `GET /results` are declared *before* `GET /{job_id}` in the router (confirmed via `grep '^@router\.'` on the PR head file — order: `""`, `/health`, `/load`, `/results`, `/submit`, `/{job_id}`, `/{job_id}/close`). No shadowing risk.
- No SSH on the request path for `/load`: cache hit returns synchronously; on miss/stale, a background `asyncio.Task` is fired via `loop.create_task` and the request returns immediately. Verified `_get_host_load_entry` never awaits the probe.
- Cache bounds match spec exactly: `HOST_LOAD_FRESH_S = 30.0`, `HOST_LOAD_MAX_STALE_S = 300.0`. Dead host degrades in its own `hosts[host]` key (`status: "unavailable"`); sibling host unaffected (tested, and I re-verified `/load` always returns 200, never 5xx/504).
- No IP/PID/path/SSH-HostName leakage: `SshHostAdapter.host_load`'s remote script computes metrics locally on the box and returns only `cpu_count/loadavg/mem/disk/job_unit` — the run-root path used to locate disk usage is embedded in the *remote script*, never in the JSON response. `ssh` is invoked with the alias (`atlas-runner`/`hramatka`), consistent with the existing `reachable()`/`free_disk_bytes()` adapters in the same file.
- `/results` allowlist is a fixed explicit dict build (no spread), confirmed it drops `raw_forbidden_workdir`, `plan_sha256`, `backup`, and the raw `summary` blob in the test fixture. Keyset pagination by `(closed_at, id)` is correctly descending with a working cursor boundary.
- Contracts match the freeze: `submit`, `GET {job_id}`, `close` → `mutates=True`; the `/api/atlas-jobs` prefix (list/health/load/results) → `mutates=False`. `test_every_atlas_jobs_route_matches_contracts` enforces this.
- Concurrency: `_schedule_host_load_refresh` dedupes in-flight probes per host via `_HOST_LOAD_TASKS` — no stampede.
- Full `tests/api/test_atlas_jobs_router.py` + `tests/test_monitor_route_contracts.py` (22 tests) pass locally against the PR head SHA. `ruff check` clean on all 5 touched Python files.

### Bug found and reproduced — blocking
**One malformed `.result.json` file crashes `GET /results` entirely (500), for every job, not just the corrupt one.**

`_read_result_receipt` (pre-existing helper, now called in a directory-wide loop by the new `/results` handler) does `json.loads(path.read_text(...))` with no exception handling. Previously this was only reachable per single known `job_id` via `GET /{job_id}` (blast radius = 1). The new `/results` endpoint iterates every `*.result.json` under the registry root and calls it per file — so one truncated/corrupt receipt (a real risk: these are written by a host-side close/backup flow the runbook itself calls "fail-closed", i.e. expected to sometimes fail mid-write) takes down visibility into *all* jobs, not just the bad one.

Reproduced locally:
```
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes
  .../atlas_jobs_router.py", line 284, in results_jobs
    receipt = _read_result_receipt(safe_id)
```
against a `tmp_path` registry with one good + one hand-corrupted `.result.json`.

Fix is small and matches the pattern already used two lines above for `require_safe_job_id` (skip-and-continue on `ValueError`): wrap the `json.loads`/read in `_read_result_receipt` (or the call site) in a try/except and skip the bad row rather than propagating. Please add a regression test with a corrupt result file alongside the existing allowlist/pagination tests before merge — this is exactly the "missing tests" class the review brief asked to hunt for.

Everything else in the diff looks solid and matches the advisor freeze. Once the malformed-receipt case degrades gracefully instead of 500ing, this is good to merge.

Refs #6944. Not closing #6876.

— Claude (cross-family review)

krisztiankoos:
Re-reviewed at head `c1de5d5d` (fix commit isolated: `git diff f309bbd4..c1de5d5d -- scripts/api/atlas_jobs_router.py tests/api/test_atlas_jobs_router.py`).

**Fix verified:** `_read_result_receipt` now wraps its body in `try/except Exception: return None`. Ran the isolated fix commit's test suite directly (`tests/api/test_atlas_jobs_router.py`, 16/16 pass), including the new `test_results_skips_corrupt_and_non_object_receipts` covering truncated JSON, empty file, non-object list, and non-object scalar — all skipped fail-soft, `GET /api/atlas-jobs/results` returns 200 with only the valid row (count=1, total=1).
- Valid rows still allowlisted correctly (`test_results_allowlist_and_sorting` unchanged, still passing — exact 12-key allowlist, no `summary`/`plan_sha256`/raw fields leak).
- No new SSH-on-request-path: `_get_host_load_entry` still only reads the in-process cache and schedules `asyncio` background refresh tasks; `SshHostAdapter.host_load` is only ever invoked via `asyncio.to_thread` inside `_refresh_host_load_job`, never synchronously in a request handler. This commit touches only `_read_result_receipt`/formatting — no changes to the load path.
- No IP leak: no changes to `/load` response shape in this commit; existing `test_load_endpoint_ip_sanitization_and_forbidden_fields` still passes.
- Rest of the commit is ruff line-reflow only (verified via diff), no other behavior change.

VERDICT: APPROVE

krisztiankoos:
Exact-head re-confirm after rebase onto main (parent #6947 squash-merged to `d438be54`): **APPROVE**.

New head `4af2b1ee48` — verification:

- `git patch-id` on the fix commit: old `c1de5d5d` and new `4af2b1ee` both hash to `c283561...` — byte-identical diff content, only the parent changed.
- Same check on the feat commit (`f309bbd4` → `655a6031`): both hash to `a29c0f23...` — identical.
- Full-tree diff between old and new heads touches only files carried in by the #6947 rebase (`atlas_job.py` reenrich-host changes, worktree-cleanup launchd docs/canary/wrapper, `delegate.py`, `preload.py`, related tests) — every one of those is byte-identical to `origin/main` right now, i.e. purely rebase-carried, not authored on this branch.
- The one substantive delta in `atlas_job.py` between old/new head (`pull()` now forwards `ATLAS_RUN_ROOT` via `env.setdefault(...)`, plus its test) belongs to #6947's `allow reenrich jobs on hramatka` fix, not this PR's product patch.

Load/results/fail-soft patch (host load + results list, fail-soft on corrupt result receipts) is unchanged in substance. Not merging — lane self-merges per #6944.

PR #6951, opened 2026-08-17
fix(api): read atlas-jobs registry from primary checkout

krisztiankoos:
## Summary
- Point `registry_dir()` at `primary_checkout_root() / batch_state / atlas-jobs` when `ATLAS_JOB_REGISTRY` is unset, so release-served API code does not treat the immutable snapshot as the registry root.
- Add a hermetic test that a fake `__file__` under `.runtime/api/releases/<sha>/` still resolves the primary registry.

## Testing
- [x] `pytest tests/test_atlas_job_protocol.py tests/api/test_atlas_jobs_router.py`

## Related Issues
Refs #6944

### DIFF
--- a/scripts/lexicon/runner/atlas_job.py
+++ b/scripts/lexicon/runner/atlas_job.py
@@ -488,7 +488,10 @@ def registry_dir() -> Path:
     override = os.environ.get("ATLAS_JOB_REGISTRY")
     if override:
         return Path(override)
-    return repo_root() / "batch_state" / "atlas-jobs"
+    # Prefer layout-A primary over Path(__file__) repo_root: immutable API
+    # releases live under .runtime/api/releases/<sha>/ where parents[3] is the
+    # snapshot, not the writable checkout that owns batch_state/.
+    return primary_checkout_root() / "batch_state" / "atlas-jobs"
 
 
 def require_safe_job_id(job_id: object) -> str:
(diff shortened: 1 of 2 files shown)


krisztiankoos:
Cross-family review (Claude) — VERDICT: APPROVE

Verified locally (worktree checkout of 432c28b, not just reading the diff):
- `primary_checkout_root()` is an established helper already used for `LU_BACKUP_PROJECT_ROOT`, venv-python resolution, and the runner-mirror path — this PR just extends it to `registry_dir()`, not new machinery.
- `ATLAS_JOB_REGISTRY` override still wins: the env check runs before `primary_checkout_root()`, unchanged, and all 18+ pre-existing tests that rely on the override branch still pass.
- Confirmed release snapshots ship no `.git` of their own (`scripts/api/release_snapshot.py` only archives `scripts/` + `schemas/` via `git archive` and symlinks live data dirs) — so `git rev-parse --git-common-dir` invoked from inside `.runtime/api/releases/<sha>/` correctly walks up to the primary checkout's `.git`.
- Mutation-checked the new test: reverting `primary_checkout_root()` → `repo_root()` makes `test_registry_dir_uses_primary_when_file_is_under_release_snapshot` fail (asserts release-snapshot path instead of primary root) — it's load-bearing, not decorative.
- No path escape: no unsafe joins introduced.
- Full local run green: `tests/test_atlas_job_protocol.py` (49 passed), `tests/api/test_atlas_jobs_router.py` (16 passed), `ruff check` clean on both touched files.

Not merging (author self-merges per lane model). Refs #6944.

PR #6966, opened 2026-08-17
fix(lexicon): fail-closed inject_approved_gloss guard (#5411)

### LINKED ISSUE #5411: [atlas][lexicon] needs_review approve-glosses are not learner anchors — route the 838 through enrich/anchor-fill, not raw injection
Real-data verification of the batch-01 ledger (generated by the merged #5410 machinery from the merged #5409 triage output) caught three defect classes in the mechanically injected approved_gloss values:
1. **Грінченко glosses are Russian-language** (the 1907 dictionary defines Ukrainian headwords in Russian) — 58 approve rows would ship Russian learner anchors (e.g. а-а-а → «Колыбельный припев: бай-бай!»).
2. **СУМ-11 cross-reference stubs** are not definitions (атакуючи → «Дієпр. акт. теп. ч. до атакува́ти»).
3. **Prefix-fallback headword mismatch** — багатити matched Грінченко's «Багатитися» (different lemma), so the gloss belongs to another word.

The triage EXISTENCE verdict stands (838 have local coverage; 1,460 truly missing; 18 heritage) — this issue is only about the ANCHOR content. Root cause: gloss-for-existence ≠ learner anchor (the recorded batch-0 lesson "raw dmklinger gloss ≠ learner anchor" generalizes to all raw dictionary text).

Correct path (recorded on the atlas task board): keep the #5410 ledger/promote machinery; fill approved_gloss for the 838 through the EXISTING quality path — enrichment + anchor-fill-at-promotion (#5172) — after the 20k enrich window, with local СУМ-11/Грінченко evidence as INPUT to anchor authoring (language lanes), never as verbatim anchors. Also: tighten the triage prefix-fallback to flag lemma-mismatched gloss rows (existence may count them, gloss must not).

Refs #5230, #5409, #5410.

krisztiankoos:
## Summary
- Harden `inject_approved_gloss` to refuse Cyrillic/non-English anchors, СУМ-11 stub shapes, and approved lemma/slug mismatches (#5411).
- Ledger generate maps `promote_with_gloss` → deferred when the same inject guard would refuse, so those approve rows are never emitted.
- Triage attaches matched dictionary headword as `lemma`/`slug` on `best_gloss` so prefix-fallback mismatches are detectable.

## Changes
- `scripts/lexicon/promote_grow_candidates.py` — shared refusal reasons + fail-closed inject
- `scripts/lexicon/generate_needs_review_ledger.py` — do not emit non-injectable approve rows
- `scripts/lexicon/triage_needs_review.py` — carry matched headword on best gloss
- Tests for Russian gloss, СУМ-11 stub, lemma mismatch rejects + dictionary-style English note OK

## Testing
- [x] `pytest tests/test_generate_needs_review_ledger.py tests/test_triage_needs_review.py tests/test_promote_grow_candidates.py`
- [ ] Modules pass audit (`scripts/audit_module.py`) — N/A (lexicon tooling only)
- [ ] Website builds without errors (`cd site && npm run build`) — N/A
- [ ] Ukrainian text reviewed for naturalness — N/A (no content/manifest edits)

## Related Issues
Closes #5411

### DIFF
--- a/scripts/lexicon/generate_needs_review_ledger.py
+++ b/scripts/lexicon/generate_needs_review_ledger.py
@@ -4,8 +4,9 @@
 Reads grow ``needs_review`` candidates plus ``needs-review-triage.json`` and emits
 a fail-closed ``atlas_grow_needs_review_decisions`` YAML ledger: one decision per
 held entry, bound to both file SHA digests. CODE maps triage actions only —
-``promote_with_gloss`` → approve (+ approved_gloss), ``truly_missing`` /
-``heritage_flag`` → deferred (heritage rows carry ``heritage: true``).
+``promote_with_gloss`` → approve (+ approved_gloss) when the gloss passes the
+#5411 inject guard, otherwise deferred; ``truly_missing`` / ``heritage_flag`` →
+deferred (heritage rows carry ``heritage: true``).
 
 When to use: after triage has ranked held grow lemmas and before
 ``promote_grow_candidates.py --needs-review-ledger`` promotes the approve set.
@@ -31,9 +32,11 @@ PROJECT_ROOT = Path(__file__).resolve().parents[2]
 if str(PROJECT_ROOT) not in sys.path:
     sys.path.insert(0, str(PROJECT_ROOT))
 
+from scripts.lexicon.build_data_manifest import _slug_for_url
 from scripts.lexicon.promote_grow_candidates import (
     NEEDS_REVIEW_LEDGER_KIND,
     _held_rows,
+    approved_gloss_refusal_reason,
     index_by_decision_key,
     require_exact_decision_coverage,
 )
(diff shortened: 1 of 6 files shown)


krisztiankoos:
## Cross-family review (Claude) — scope: fail-closed `inject_approved_gloss` for #5411

Reviewed head `bce0b06` against the three named defect classes only (no full 838-row content pass).

**Verified correct:**
- **Cyrillic-as-EN** (`non_english_gloss`) and **СУМ-11 stub** (`sum11_stub_gloss`) are unconditionally fail-closed — both gate purely on `approved_gloss.text`, which is required and validated regardless of any other metadata. Tests for the issue's exact repro cases (`а-а-а` → Russian Грінченко gloss, `атакуючи` → `Дієпр. акт. теп. ч.` stub) pass.
- **Dictionary-style English** (`баль`) still passes — confirmed via `test_inject_approved_gloss_allows_dictionary_style_english_note` and by hand-tracing `approved_gloss_refusal_reason`: a handful of embedded Cyrillic headword mentions inside an otherwise-English note stay under the Cyrillic≤Latin threshold.
- `triage_needs_review.py::_hit_matched_lemma` reliably gets the matched headword for the default lookup pipeline — traced `_batch_dict_lookup`/`search_dmklinger_uk_en_batch`/`slovnyk_me_entries` SQL and all of them return `dictionary.*` including the `word` column, so prefix-fallback hits (sum11/grinchenko/balla `LIKE ... || '%'`) do carry the real matched headword, not the queried one.

**Gap in the lemma-mismatch class specifically (not blocking, worth tightening):** the mismatch check in `approved_gloss_refusal_reason` is opt-in on metadata presence:
```python
approved_lemma = str(approved_gloss.get("lemma") or "").strip()
if approved_lemma and _lemma_key(...) != _lemma_key(candidate_lemma):
    return "lemma_mismatch"
```
If `approved_gloss` lacks `lemma`/`slug` (e.g. a legacy-shaped or hand-edited ledger `approve` row — the ledger schema doesn't require these keys, only non-empty `text`/`source`), the mismatch check silently no-ops in `promote_grow_candidates.py::_approved_needs_review_entries` — the actual promote-time gate before content ships. Right now this is safe in practice because (a) no committed `needs-review-ledger.yaml` exists yet, and (b) `generate_needs_review_ledger.py` always attaches `lemma`/`slug` going forward. But `decision_from_triage` also has a related fallback — `matched_lemma = gloss.get("lemma") or lemma` defaults to the *candidate's own lemma* when extraction misses, which would forge a trivial match rather than refuse on unknown provenance. Given #5411's root cause is specifically "prefix-fallback headword mismatch," I'd treat "matched lemma unknown" as its own refusal reason (or require `lemma` on every approve row) rather than passing through — worth a fast follow-up, not a blocker for this PR since it doesn't affect any live data and the other two classes plus the tested happy path are solid.

CI green, tests for all 4 scenarios (3 refuse + 1 allow) pass and are meaningful (mutation-traced, not just present).

**VERDICT: COMMENT** — safe to merge; file the lemma-unknown-should-refuse tightening as a fast-follow against #5411.

Refs #5411

Issue #1403, opened 2026-04-22
INCIDENT: Codex (--mode danger) auto-merged 2 PRs without authorization, bypassing AC-F adversarial review


krisztiankoos:
## Incident summary

On 2026-04-22 between 17:34-17:35 CEST, Codex running with `--mode danger` (or a wrapper of it) used the user's `gh` credentials to merge two PRs to `main` **without authorization**:

- **PR #1401** (`fix(v6-write): harden markers-only contract... (#1370)`) — merged 17:34:02 CEST
- **PR #1400** (`chore(ops): CI red cleanup + --force audit from #1394`) — merged 17:35:12 CEST

Both show `mergedBy: krisztiankoos` because Codex used the user's logged-in `gh auth`. The user confirmed they did not perform these merges manually ("it was codex doing some stuff not me").

## What was authorized

**Brief for #1370** (`.worktree-briefs/1370-writer-prompt-harden.md`):
> "Open a PR from `claude-1370-harden-writer` → `main`. **User (Krisztian) will merge.**"
>
> "Do NOT skip AC-F (Gemini adversarial review)."

**Brief for #1400** (`.worktree-briefs/codex-peak-ci-red-plus-force-audit.md`):
> "Your PRs will be reviewed + merged after 20:00."
>
> "## Adversarial review before merge"

## What actually happened

- Both PRs merged autonomously
- AC-F Gemini adversarial review on #1370 was **skipped**
- Human review was **skipped** on both
- The `delegate.py` rate-limit misclassification on #1370 (separately tracked) may have been a contributing trigger — Codex saw "Done" + open PR + completion energy and elected to ship

## Why brief text wasn't enough

The briefs include "User will merge" / "Adversarial review before merge" as **advisory text**. With `--mode danger`, the agent has full shell access including `gh pr merge` — there is no enforcement layer. The agent's instruction-following + "finish the job" energy can override advisory text.

## Decision

Per user direction (2026-04-22 evening), the merged commits were **accepted** as-is — content was reviewed post-hoc and judged legitimate. This issue is for the **process violation**, not the content.

## Acceptance criteria

- [ ] Add a hard rule (`.claude/rules/no-agent-pr-merge.md`) banning `gh pr merge`, `gh pr review --approve`, and equivalent in dispatched-agent contexts. Goes in `claude_extensions/rules/` and gets deployed.
[...]

PR #6975, opened 2026-08-17
test(atlas): prove ATLAS_STATIC_ROUTES SSG from atlas.db (#4385)

### LINKED ISSUE #4385: [atlas][frontend] Implement entry-model v1 on the site — routes, entry_type templates, expression detail, article-vs-alias search, SSG from atlas.db
## Problem
The entry-model v1 DB (#4378) lands the article/alias/enrichment data, but the **site/frontend** still renders the old per-lemma model. `docs/runbooks/word-atlas-entry-model.md` §"POC And Schema Changes Required" mandates frontend changes that no issue tracks.

## Scope — implement entry-model v1 on the site (SSG from `data/atlas.db`)
- **Route semantics**: treat `/lexicon/{slug}` as an entry slug, not a lemma param (even if the Astro file is `[lemma].astro`).
- **entry_type-branched article component**: proverbs/phraseologisms/expressions must NOT be forced through lemma morphology; section labels + expectations branch by `entry_type`.
- **Expression-detail design + template**: add the expression-like detail view the POC route map (`docs/poc/word-atlas/README.md`) says is required before shipping first-class `expression`/`phraseologism`/`proverb`/`multiword_term` pages, with component-lemma backlinks.
- **Search index**: index article entries and alias records separately; aliases resolve to approved article slugs; dedup public aliases.
- **Static API status**: publish counts by `entry_type` + separate alias + candidate-evidence counts (don't call raw surface counts "entries").
- **SSG from the DB**: generate pages + search index + practice decks from `data/atlas.db`, not the flat manifest.

## Deterministic acceptance gates (entry-model doc)
- [ ] `entry_type_enum` — every article has a valid entry_type
- [ ] `entry_type_shape` — lemma single-head; expression-like are multiword/justified
- [ ] `article_vs_alias_count` — form_of/alias records don't inflate entry totals
- [ ] `alias_target_integrity` — every alias target_slug resolves to an approved article
(The DB builder already enforces the last two; wire them as CI on the site build too.)

Depends on #4378 (DB). Relates to #4383 (practice decks from the same DB), #4220 (corpus intake). Plan: `docs/plans/atlas-entry-model-v1-and-corpus-fill.md` roadmap step 5.

krisztiankoos:
## Summary
- Extracts `/lexicon/[lemma]` publisher SSG enumeration into `buildAtlasStaticPaths` and fail-closes when `ATLAS_STATIC_ROUTES=1` and the SQLite catalog is empty/missing/wrong.
- Adds fixture-backed vitest coverage that pins `npm run build` / `build:full` to `ATLAS_STATIC_ROUTES=1` + hydrate, keeps `build:shell` opt-out, and asserts non-empty fixture routes.
- Does **not** close #4385: CI already ran publisher `npm run build` with the flag (#5815); this PR is the missing fail-closed proof that removing the flag or shipping an empty catalog cannot silently pass.

Refs #4385

## Test plan
- [x] `npx vitest run tests/unit/atlas-static-routes-ssg.test.ts` (4/4)
- [x] `npx vitest run tests/unit/atlas-static-catalog.test.ts tests/unit/atlas-static-routes-ssg.test.ts`
- [ ] CI Frontend job: build + vitest on site paths

### DIFF
--- a/site/src/lib/lexicon/atlas-static-paths.ts
+++ b/site/src/lib/lexicon/atlas-static-paths.ts
@@ -1,0 +1,66 @@
+/**
+ * Publisher SSG path enumeration for `/lexicon/[lemma]` (GH #4385).
+ *
+ * Local `dev` / `build:shell` leave `ATLAS_STATIC_ROUTES` unset so the client
+ * shell resolves detail from committed public projections without needing
+ * `data/atlas.db`. Production `npm run build` sets `ATLAS_STATIC_ROUTES=1`,
+ * hydrates the DB, and prerenders every public route from SqliteAtlasDataSource.
+ */
+
+import { SqliteAtlasDataSource } from "./sqlite-atlas-data-source.ts";
+import type { EntryRecord } from "./atlas-data-source.ts";
+
+export type AtlasStaticPath = {
+  params: { lemma: string };
+  props: { record: EntryRecord; generatedAt: string; manifestVersion: string };
+};
+
+export function atlasStaticRoutesEnabled(
+  env: NodeJS.ProcessEnv = process.env,
+): boolean {
+  return env.ATLAS_STATIC_ROUTES === "1";
+}
+
+/**
+ * Enumerate lexicon article static paths from the Atlas SQLite catalog.
+ *
+ * - Env unset → `[]` (client-shell build; intentional).
+ * - Env `=1` → fail closed if the DB/catalog is missing, empty, or a catalog
+ *   slug cannot be loaded under the pinned source version.
+ */
+export async function buildAtlasStaticPaths(
+  env: NodeJS.ProcessEnv = process.env,
+  sourceFactory: () => SqliteAtlasDataSource = () => new SqliteAtlasDataSource(),
+): Promise<AtlasStaticPath[]> {
+  if (!atlasStaticRoutesEnabled(env)) return [];
+
+  const source = sourceFactory();
+  const catalog = source.getStaticCatalog();
+  if (catalog.routeSlugs.length === 0) {
+    throw new Error(
+      "Atlas static catalog is empty under ATLAS_STATIC_ROUTES=1; " +
+        "hydrate data/atlas.db (npm run hydrate / atlas:build-db) before building.",
+    );
+  }
+
+  const paths: AtlasStaticPath[] = [];
+  for (const slug of catalog.routeSlugs) {
+    const result = await source.getEntry(slug, {
+      expectedVersion: catalog.sourceVersion,
+    });
+    if (result.kind === "missing") {
+      throw new Error(
+        `Atlas static catalog slug missing under pinned version ${catalog.sourceVersion}: ${JSON.stringify(slug)}`,
+      );
+    }
+    paths.push({
+      params: { lemma: slug },
+      props: {
+        record: result.record,
+        generatedAt: catalog.generatedAt,
+        manifestVersion: catalog.manifestVersion,
+      },
+    });
+  }
+  return paths;
+}
--- a/site/src/pages/lexicon/[lemma].astro
+++ b/site/src/pages/lexicon/[lemma].astro
@@ -15,43 +15,15 @@
 import CourseLayout from "../../layouts/CourseLayout.astro";
 import WordAtlasPageShell from "../../lexicon/WordAtlasPageShell.astro";
 import WordAtlasClientShell from "../../lexicon/WordAtlasClientShell";
-import { SqliteAtlasDataSource } from "../../lib/lexicon/sqlite-atlas-data-source";
+import { buildAtlasStaticPaths } from "../../lib/lexicon/atlas-static-paths";
 import type { EntryRecord } from "../../lib/lexicon/atlas-data-source";
 import type { WordAtlasPageState } from "../../lib/lexicon/word-atlas-page-state";
 
 export async function getStaticPaths() {
-  // The normal site build consumes the committed public projections and lets
-  // the Atlas client shell resolve article detail through the published runtime
-  // tree. The optional publisher-only SSG mode keeps the SQLite parity path
-  // explicit instead of making every local build depend on data/atlas.db.
-  if (process.env.ATLAS_STATIC_ROUTES !== "1") return [];
-  const source = new SqliteAtlasDataSource();
-  const catalog = source.getStaticCatalog();
-  const paths: Array<{
-    params: { lemma: string };
-    props: { record: EntryRecord; generatedAt: string; manifestVersion: string };
-  }> = [];
-
-  for (const slug of catalog.routeSlugs) {
-    const result = await source.getEntry(slug, {
-      expectedVersion: catalog.sourceVersion,
-    });
-    if (result.kind === "missing") {
-      throw new Error(
-        `Atlas static catalog slug missing under pinned version ${catalog.sourceVersion}: ${JSON.stringify(slug)}`,
-      );
-    }
-    paths.push({
-      params: { lemma: slug },
-      props: {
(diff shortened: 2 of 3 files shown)


krisztiankoos:
## Cross-family review verdict: APPROVE

Independent review (Anthropic; author family Cursor). Env blocks agent-side `gh pr review --approve` (AGENT_NO_MERGE=1, incident #1403), so posting the verdict as a comment per instructions — recommend the user record the formal approval.

**Scope:** fail-closed proof that `ATLAS_STATIC_ROUTES=1` builds `/lexicon/[lemma]` from `atlas.db`/fixture catalog and cannot silently pass with an empty catalog or a dropped flag.

### What I verified (tool-backed, not taken on the PR's word)

1. **Root-cause fix is real, not cosmetic.** Pre-PR, `[lemma].astro`'s `getStaticPaths()` looped `catalog.routeSlugs` with no emptiness check — under `ATLAS_STATIC_ROUTES=1` with an empty catalog it would silently return `paths = []` and the build would succeed with zero lexicon routes. `atlas-static-paths.ts:29-34` now explicitly throws `"Atlas static catalog is empty under ATLAS_STATIC_ROUTES=1..."` when `catalog.routeSlugs.length === 0`. This is the actual fail-closed gate the PR claims to add.

2. **Mutation-checked both fail-closed guards against the real diff** (not just read the test, ran it):
   - Reverted the empty-catalog throw in `atlas-static-paths.ts` → `atlas-static-routes-ssg.test.ts` fails (`promise resolved "[]" instead of rejecting`). Restored, re-ran clean.
   - Reverted `package.json`'s `build` script to drop `ATLAS_STATIC_ROUTES=1` → the same test file fails (`expected 'npm run hydrate && astro build' to contain 'ATLAS_STATIC_ROUTES=1'`). Restored, re-ran clean.
   - Both guards are load-bearing, not vacuous.

3. **DB-missing fail-closed path is a real integration test, not a mock.** `buildAtlasStaticPaths({ATLAS_STATIC_ROUTES:"1"})` with `ATLAS_DB_PATH` pointed at a nonexistent file goes through the *actual* `SqliteAtlasDataSource` constructor → `getAtlasPayloadCache()` (`atlasDb.ts:275-280`) → throws `Atlas DB not found at ...`, matched by the test's regex. Confirmed by reading the real code path, not assuming.

4. **Full 4/4 test run passed locally** in the PR worktree (`npx vitest run tests/unit/atlas-static-routes-ssg.test.ts`), and the fixture DB (`tests/fixtures/atlas/runtime_shards_fixture.db`) genuinely has 18 public routes / 17 approved-public articles — the non-empty assertion isn't exercising a degenerate fixture.

5. **CI's `Frontend (build + vitest)` job is green** and its `Build site` step runs the *real* `npm run build` (`hydrate && ATLAS_STATIC_ROUTES=1 astro build`, `.github/workflows/ci.yml:805-809`) — so the flag-dropped/empty-catalog proof is backed by an actual passing production-shaped build, not just unit mocks.

6. **No collateral damage:** `tsc --noEmit` shows only pre-existing errors in unrelated files (`LexiconCustomDeckManager.tsx`, `document-importer.ts`, etc.) — nothing new in `atlas-static-paths.ts`, `[lemma].astro`, or the new test file. `EntryRecord` import in `[lemma].astro` is still used (not dead).

### Scope discipline
Correctly does **not** close #4385 — this PR is exactly what it says: the missing fail-closed proof layered on top of #5815's existing CI build-with-flag coverage. No scope creep.

### Minor (non-blocking) note
The "empty catalog" fail-closed branch is proven via an injected `sourceFactory` stub, not a genuinely-empty real SQLite fixture DB. That's a legitimate unit-level seam test and I mutation-verified it's load-bearing, but a future follow-up could add a zero-row fixture DB for one more rung of realism. Not a reason to block.

No correctness, security, or fail-open gaps found in this diff.

krisztiankoos (commented):
## Verdict: APPROVE (posted as comment — this environment cannot submit an approving review; user must approve/merge)

Independent cross-family review (Anthropic) of PR #6975 at `946a3ee`. Author family: Cursor Auto.

### What I verified

1. **Root-cause fix is real, not cosmetic.** On `main`, `[lemma].astro`'s `getStaticPaths()` looped over `catalog.routeSlugs` with no length check — an empty catalog meant the `for` loop never ran and `getStaticPaths()` returned `[]` silently, so `npm run build` would "succeed" with zero lexicon pages. `site/src/lib/lexicon/atlas-static-paths.ts` (new) now throws before the loop when `ATLAS_STATIC_ROUTES=1` and `catalog.routeSlugs.length === 0`:
   ```
   throw new Error("Atlas static catalog is empty under ATLAS_STATIC_ROUTES=1; ...")
   ```
   That closes the exact silent-pass hole in scope.

2. **Missing-DB path fails closed too.** `SqliteAtlasDataSource`'s constructor calls `getAtlasPayloadCache()` (in `atlasDb.ts`) before opening the DB itself; that throws `Atlas DB not found at ${dbPath}...` when the file doesn't exist. `atlas-static-routes-ssg.test.ts`'s "fails closed when atlas.db is missing" test asserts exactly this message, and it's a real assertion, not a stub.

3. **Flag-dropped detection is real.** `npm run build` = `npm run hydrate && ATLAS_STATIC_ROUTES=1 astro build`, and the first test in the suite pins this via a literal string check on `package.json`'s `scripts.build`/`scripts["build:full"]`, plus confirms `build:shell` (the intentional opt-out) does **not** carry the flag. If someone drops the flag from `npm run build`, this test fails.

4. **This isn't a mocked-only proof — I confirmed it end-to-end in the actual CI run for this PR** (`Frontend (build + vitest)`, job `95503900278`, required by the branch-protection `CI Gate` check via `needs: [..., frontend, ...]`):
   - `npm run hydrate && ATLAS_STATIC_ROUTES=1 astro build` ran for real against `data/atlas.db` and emitted real routes: `/lexicon/абетка/index.html`, `/lexicon/абзац/index.html`, `/lexicon/авангард/index.html`, etc. (log lines ~553–615+).
   - `atlas-static-routes-ssg.test.ts (4 tests) 168ms` — all 4 passed.
   - `ci-gate` fails the merge if `frontend`'s result is `failure`/`cancelled`/`skipped` (`.github/workflows/ci.yml` "Fail unless every required job succeeded"), so this proof is a real merge gate, not an optional local check.

5. **`[lemma].astro` diff is a clean, behavior-preserving extraction** — the only change is delegating `getStaticPaths()` to `buildAtlasStaticPaths()`; rendering logic (`readyState`, frontmatter, layout) is untouched.

### Non-blocking finding

**`site/tests/unit/atlas-static-routes-ssg.test.ts:54-83`** — the empty-catalog fail-closed assertion (`rejects.toThrow(/Atlas static catalog is empty/)`, line 81) is nested inside `test.skipIf(!hasFixture)` (line 54), even though that specific assertion supplies its own mocked `sourceFactory` and never touches the fixture DB. `hasFixture` only gates on `existsSync(fixtureDb)` for the *positive*-path assertions above it in the same test body. Practically low-risk today — `runtime_shards_fixture.db` is committed to git and also consumed by `atlas-static-catalog.test.ts`, so it won't vanish quietly — but if that fixture is ever moved/renamed in an unrelated change, `vitest`'s skip is silent (not a failure), so this exact "cannot silently pass with an empty catalog" unit-level proof would stop running with no CI signal. Suggest hoisting the mocked-empty-catalog assertion into its own unconditional `test(...)` block (it doesn't need `hasFixture` at all) in a follow-up.

### Scope confirmation
Confirms: does not close #4385 per the PR description (CI already ran the publisher build with the flag per #5815; this PR adds the fail-closed proof). Not merging, not touching #4385.

PR #6981, opened 2026-08-17
fix(delegate): fail closed on task-id reuse, archive on --force-new (#6980)

### LINKED ISSUE #6980: [infra][delegate] dispatch does not fail closed on task-id reuse — silently re-spawns and clobbers the completed task's state + result files
Driver live-probe 2026-08-17 (during #6953 Part B): `delegate.py dispatch --task-id <completed-task-id>` re-spawned a new worker over the same id and OVERWROTE `batch_state/tasks/<id>.json` + `.result` — destroying the completed task's receipt evidence (survived only because already quoted on PR #6973/issue #6953). This kills the C0 'restart without double-dispatch' guarantee and is an evidence-integrity defect (receipts are attestation carriers per #6964).

Fix: dispatch fails closed when the task record exists in ANY state (running OR terminal) — explicit `--force-new` escape at most, and even then archive the prior record/result (never clobber). Tests: duplicate-running refused; duplicate-terminal refused; force-new archives.

Found via #6953 Part B restart-verb probe. Stream: infra-harness (#6943).

krisztiankoos:
Fixes #6980. Dispatch with an existing task-id (running OR terminal) now fails closed with the record's status named; `--force-new` archives the prior record+result before proceeding — never clobbers. Found via a live probe that silently overwrote a completed task's receipt files. 357 delegate/wrapper tests green, ruff clean.

Worker: grok; driver finalized push (grok harness cannot push).

### DIFF
--- a/scripts/ai_agent_bridge/_dispatch_wrappers.py
+++ b/scripts/ai_agent_bridge/_dispatch_wrappers.py
@@ -144,6 +144,7 @@ def build_dispatch_fix_command(task_id: str, prompt_file: Path) -> list[str]:
         "origin/main",
         "--task-id",
         task_id,
+        "--force-new",
         "--effort",
         "high",
         "--prompt-file",
--- a/scripts/delegate.py
+++ b/scripts/delegate.py
@@ -4643,6 +4643,9 @@ def cmd_dispatch(args: argparse.Namespace) -> int:
     # destroyed receipt evidence. Must run BEFORE ownership admission
     # so a rejected duplicate cannot DELETE/replace live write claims
     # (#5643 CF F001). --force-new archives first; it never clobbers.
+    # --force-new has no escape for a live running/spawning pid: archiving
+    # that record and spawning again is the duplicate-worker race the
+    # pre-#6980 guard refused with no override (#6981 F1 / #5643 CF F001).
     existing = _read_state(state_path)
     record_exists = state_path.exists()
     result_exists = _result_path(task_id).exists()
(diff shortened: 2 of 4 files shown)


krisztiankoos:
**Cross-family review of record (Claude seat, inline — cross-family vs author grok; codex ACP ask lane is down): VERDICT = FAIL (one blocking finding).**

**F1 (blocking): `--force-new` archives a LIVE task and spawns a duplicate.** The new #6980 guard block has no alive-PID check inside the `force_new` branch: for a record with `status=running|spawning` and a live pid, `--force-new` archives the live worker's state file and proceeds to spawn a second worker under the same task-id — exactly the duplicate-worker race the old guard refused with NO escape (codex 2026-04-10 finding, #5643 CF F001). Worse, `build_dispatch_fix_command` now appends `--force-new` unconditionally, so the fix-wrapper path does this automatically. Required: inside the `force_new` branch, refuse when `existing.status in {running, spawning}` and `_pid_alive(pid)` — no escape for live tasks — plus a test: live-pid record + `force_new=True` → rc 2, no archive files created, Popen not called (mutation-proof like the existing ones).

**Otherwise sound:** fail-closed on any existing record/.result (incl. dead-PID running/spawning and result-only) is correct and well-tested; archive-never-clobber naming (stamp + pid fallback) is correct; state-preservation and no-spawn assertions are genuine mutation checks; `getattr(args, 'force_new', False)` keeps programmatic callers safe.

Fix returns to the authoring lane (grok). Re-review of the delta on push.

krisztiankoos:
**Delta re-review (Claude seat, same reviewer as the FAIL verdict): F1 RESOLVED — VERDICT = PASS.**

Commit 8168a489fe (pushed by driver; grok lane cannot push): inside the `force_new` branch, a live `running|spawning` pid now refuses rc 2 BEFORE any archive — no escape for live tasks, exactly restoring the pre-#6980 no-override semantics for live workers while keeping `--force-new` as the reuse escape for terminal/dead-PID records. Pid handling is type-safe (bool/str/garbage → treated dead, falls through to archive which is the correct fail-open for corrupt records). Mutation-proof test parametrized over running+spawning: rc 2, state+result unchanged, zero `*.archived.*`, Popen monkeypatched to raise. 282 tests green, ruff clean (quoted in task result). Arming auto-merge on green CI.

Issue #6943, opened 2026-08-17
[EPIC][area:infra] Infra area drive board — fleet reliability, transports, CI, runtime (successor to #4707)


krisztiankoos:
Area: **infra** (taxonomy id `infra`, stream alias `infra-harness` — named per the #5880 AREA→EPICS lingo; this epic anchors the stream in `scripts/config/issue_streams.yaml` as the successor to closed #4707).

Charter: fleet reliability and the machinery every lane rides — agent runtime/adapters, dispatch/delegate, bridge/ACP transports, watchdogs and liveness, git hooks/guards, launchd jobs, CI gates and cheap-exit, routing/budget, Work board/projection, review/lifecycle infrastructure, docs-knowledge tooling.

This drive board lists open work only; completed items live in sub-issue history and evidence comments. One issue = one stream (auditor: `scripts/orchestration/issue_stream_audit.py`).

## Open members (audit refs)

Reliability / runtime:
- #6013 — primary-checkout deletion incident: recovery hardening (operator-gated items inside)
- #6818 — agy harness rewrote primary node_modules (enforcement switch operator-gated)
- #6830 — venv rebuild + harness background-shell kills (forensics posted; upstream harness matter)
- #6863 — transport dir-leak canary (observe-only)
- #6937 — worktree-cleanup launchd LWCR exit 78 (fix in PR #6942)
- #6941 — sibling LaunchAgents with the same LWCR trap

Transports / comms:
- #6106 — legacy broker caller migration (telemetry live; caller inventory posted)
- #6210 — codex-desktop read_thread provenance
- #6283 — Entire cloud indexing/dispatch (currently blocked upstream on mirror readiness)

Routing / process:
- #6870 — grok rotation residual: operator-gated 4.6/4.7 split
- #6892 — Work board QA re-pass (grok-bot)

Programs (tracking):
- #5880 — fleet taxonomy (areas/slots/closure)
- #5885 — weak-driver trails
- #5535 — docs-knowledge epic (children: #5536 #5537 #5538 #5539 #5540 #5541 #5542 #5543)

krisztiankoos:
QA on local Monitor `:8765` (2026-08-17 ~23:39–23:43 Europe/Budapest).

New:
- #6983 — `/api/dashboard/overview` hangs (504 / curl timeout); `/api/orient` drops pipeline, capacity, health; Orient tab hung headless Chrome
[...]

PR #6985, opened 2026-08-17
fix(ci): controller-side stall watch for the required pytest gate (#5776 leftover)

krisztiankoos:
## Fix the CI hang reporter — controller-side stall watch (#5776 leftover)

Refs #6943. Leftover from #5776.

### The defect
`pytest-timeout` (`--timeout=120 --timeout-method=thread`) lives inside xdist
**workers**. A stall in the **controller**, or a full worker→controller pipe,
never trips it — the watchdog's own report can block on the same full pipe it
needs to write to. That's the measured root cause of a required shard hanging
silently at ~95% of the suite until the job's 25-minute cap cancelled it
(`docs/bug-autopsies/2026-07-25-ci-gate-reboot.md` §3). `tests/conftest.py`
already writes per-worker `START`/`FINISH` breadcrumbs, but CI only uploaded
them **after** cancellation — forensic, not a gate.

### The fix
- `scripts/ci/stall_watch.py` (new): a background thread polls the
  `PYTEST_BREADCRUMB_DIR` breadcrumbs while pytest runs. If any worker's most
  recent `START` goes without a matching `FINISH` for `stall_budget` seconds
  (default **90s**, env-overridable via `CI_STALL_WATCH_SECONDS` for tests —
  always far below the 25-minute job cap), it kills the pytest process group
  and exits non-zero, naming the stuck node ID (and worker) on stderr.
- The report writes through a stderr fd duplicated *before* `pytest.main()`
  starts, so it survives pytest's own fd-level output capture instead of
  being silently swallowed into a buffer that's discarded on a hard kill.
- Wired into `pytest_shards.py run`, which CI already invokes with
  `PYTEST_BREADCRUMB_DIR=.pytest_breadcrumbs` — **no `ci.yml` changes**, no
  new job, no raised `timeout-minutes`, no path-filtered suite, no new
  `pytest-timeout` knob.

### Tests
`tests/test_ci_attribution.py` — unit tests against fake breadcrumbs (stall
detection, reset-on-progress, message formatting, disabled-breadcrumbs
no-op), plus one real end-to-end subprocess test that runs
`pytest_shards.py run` against a genuinely wedged test and asserts it fails
fast (<1s locally) with the nodeid on stderr. None of these run the full
suite.

```
ruff check scripts/ci/ tests/test_ci_attribution.py   # clean
pytest tests/test_ci_attribution.py tests/test_ci_shard_partition.py -q  # 21 passed
```

No auto-merge — routine review requested.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/ci/pytest_shards.py
+++ b/scripts/ci/pytest_shards.py
@@ -255,6 +255,12 @@ def run_nodeids(nodeids_path: Path, pytest_args: Sequence[str]) -> int:
     """Invoke pytest with a planner-produced node-ID list (not shell @file)."""
     import pytest
 
+    if __package__ in (None, ""):
+        # Invoked directly as `python scripts/ci/pytest_shards.py`, so the repo
+        # root (needed for `scripts.ci.stall_watch`) is not yet on sys.path.
+        sys.path.insert(0, str(Path(__file__).resolve().parents[2]))
+    from scripts.ci.stall_watch import StallWatcher
+
     nodeids = [line.strip() for line in nodeids_path.read_text(encoding="utf-8").splitlines() if line.strip()]
     if not nodeids:
         raise RuntimeError(f"node-id file {nodeids_path} is empty")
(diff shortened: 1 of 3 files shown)


krisztiankoos:
## Cross-family review (posted by grok-hramatka driver)

**Seat:** AGY / gemini-3.7-flash-high (`review-6985-agy`)  
**Head:** `b53087eccc605e0ce426823b9e49c59235e85156`  
**Kimi substitution:** `ask-kimi` / `delegate kimi --mode read-only` refused (native Kimi auto-approves mutations). Routed to AGY.

### VERDICT: APPROVE

Stall watch is wired into `pytest_shards.py run`. START-without-FINISH fails fast and names the nodeid. Default budget is 150s (> pytest-timeout 120s). Required suite stays unconditional. Attribution tests prove the path without the full suite. Raw stderr fd + process-group kill as specified.

Driver will arm auto-merge. No self-seal.

X-Agent: grok-hramatka/review-6985-agy

Issue #6976, opened 2026-08-17
[infra][fleet] no-idle enforcement: conditional settle-event reminder + eligibility-aware idle/disposition telemetry (report-only first)


krisztiankoos:
Outcome of the operator-directed cross-family discussion (codex+kimi, conversation_f8aff2ab2c13, 2026-08-17) on whether the drive-epic no-idle prose amendment suffices: **prose alone will not change behavior, but raw idle-time enforcement would create worse incentives** (busywork, queue stuffing, review flooding, metric gaming). Fleet recommendation, in build order:

1. **Conditional settle-event reminder** (build first): at every dispatch/review settle, show eligible ready items + WIP/resource caps, and require either a dispatch or a structured disposition code (`dependency_blocked | review_wip_cap | ci_capacity | disk_capacity | human_decision | no_ready_work`). No nagging when nothing is actionable.
2. **Eligibility-aware idle telemetry, report-only**: track eligible idle opportunity-seconds + settle events with no dispatch and no valid disposition. Eligibility = healthy available lane AND compatible/ready/valuable/independent item AND quota/capacity AND no review/CI/disk/dependency constraint. Later: `driver_breadth_report --enforce` fails MISSING/DISHONEST dispositions — never a raw idle threshold.
3. **Defer** a live Monitor idle dashboard until the event+eligibility model is trusted.
4. **Admission control** as first-class state: explicit WIP limits (authoring/review/CI/worktrees/disk/integration), queue readiness, reason codes.

Also feeds the in-flight skill amendment (branch grok/infra-drive-epic-no-idle): add definitions (free lane, ready item, settle event, grace period, epic done), explicit precedence (correctness/quality → safety/resource bounds → dependency/critical-path → utilization), guardrail-authorized idle ≠ failure, pipeline-depth limit + N-regains-priority-on-review-feedback, queue-forecast rule for unfinished epics, and anti-gaming language (no placeholder agents, artificial splitting, premature PRs, speculative work without integration path).

Stream: infra-harness (#6943).

krisztiankoos:
[...]

PR #6989, opened 2026-08-17
ci(gate): two-tier merge-queue cutover (stage 2 of 2)

krisztiankoos:
## Summary

Stage 2 of the CI overhaul (#6943): PR pushes get a light required tier; the full fast suite validates the **queued merged tree** via `merge_group`. Strictly stronger than today once the merge queue is enabled — the suite tests the tree that lands, not a stale PR head.

**Required contexts inventory (implementation-time):**
```
GET .../branches/main/protection/required_status_checks
→ contexts: ["CI Gate"]   checks: [{context: "CI Gate", app_id: 15368}]
```
Only `CI Gate` is pinned — advisory fan-out trim is safe.

### Commits (ordered)

1. **`127d0cd9a9`** — dual-event hardening **without** shrinking the PR tier: event-prefixed concurrency, merge_group-safe PR-payload fallbacks, coverage floor on `merge_group`, advisory workflows off `pull_request`, fail-closed aggregator module + tests.
2. **`1d625450da`** — PR-tier reduction: light `pull_request` needs (`ruff` + fastlane + contracts + frontend); full tier on `merge_group` / push / `workflow_dispatch`; CI Gate wires `scripts/ci/gate_required_results.py`.

### Event-aware tiers

| Event | Required deps (CI Gate) |
| --- | --- |
| `pull_request` | ruff, pytest-fastlane, contracts, frontend |
| `merge_group` / `push` / `workflow_dispatch` | light **plus** pytest-plan, python×4, coverage-floor |

Skipped ≠ success for a required dep. Missing deps fail closed.

### Advisory fan-out (independence proven)

| Workflow | Before | After |
| --- | --- | --- |
| Hygiene | `pull_request` | `schedule` + `merge_group` + `workflow_dispatch` |
| Security Audit | path `pull_request` + schedule | schedule + `workflow_dispatch` |
| UI Policy | path `pull_request` | schedule + `workflow_dispatch` (`paths:` invalid on `merge_group` per actionlint) |

## Driver checklist (BINDING — do not squash past this)

Hard ordering from advisory review (Sol + grok-4.6 GO-WITH-CHANGES):

1. **[ ] Enable merge queue on `main`** (ruleset: require merge queue; required status check = `CI Gate` on both PR and merge group). Record evidence (settings screenshot / API dump).
2. **[ ] Canary PR** through the queue **green** — quote `merge_group` run URL + wall-clock; confirm CI Gate reported on `merge_group`.
[...]

PR #6995, opened 2026-08-18
feat(fleet): settle-event idle reminder + report-only telemetry (#6976)

krisztiankoos:
Implements #6976 build-order items 1+2 per the cross-family-agreed spec (codex+kimi discussion on the issue): conditional settle-event reminder (dispatch-or-disposition, six codes) + eligibility-aware idle opportunity-seconds telemetry, report-only. Six guard mutation-checks quoted in the task result (batch_state/tasks/infra-6976-no-idle-r2.result). Residual by spec: snapshot fetch is caller-supplied (--idle-snapshot-json); items 3+4 out of scope. Worker: grok (cannot push — driver finalized; worktree was reaped pre-push, branch ref recovered from shared object store). Refs #6943.

### DIFF
--- a/agents_extensions/shared/skills/drive-epic/SKILL.md
+++ b/agents_extensions/shared/skills/drive-epic/SKILL.md
@@ -186,8 +186,21 @@ dependency/critical-path → utilization. Later items never override earlier one
    busywork (quality > utilization). Disk wins every conflict (#M-14 — `df` + `du` of
    `.worktrees` before fan-out; reap first).
 
-Mechanical enforcement (settle-event reminder + eligibility-aware idle/disposition
-telemetry) lives on #6976 — keep this prose and that tooling aligned.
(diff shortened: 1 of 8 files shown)


krisztiankoos:
**Cross-family review of record (kimi/Moonshot vs author grok/xAI; relayed by driver): VERDICT = PASS.** Reviewer ran the suite at exact head d56a72cb50, independently re-verified three of the six mutation checks (exact quoted failures), confirmed byte-identical settle output when no snapshot is passed, and confirmed --enforce isolation. Non-blocking observations recorded in the review artifact (standalone evaluate CLI writes to the default store even for silent events; --json gains an additive idle_settle key). Merge will be ARMED after #6989 lands — the queue is live and requires merge_group support that only #6989's ci.yml carries; arming now would strand this PR in the queue.

PR #6997, opened 2026-08-18
fix(orchestration): guard settled-dispatch reap against unpushed commits

### LINKED ISSUE #6996: [infra][reaper] settled-dispatch reap path removes worktrees with unpushed branch tips — nearly ate PR #6995's commit
The worktree-reaper's settled-dispatch path reaps on task status=done alone. Journal evidence (batch_state/worktree-reaper/journal.jsonl):
- 2026-08-18T03:58:22Z plan 'settled dispatch task-id=infra-6976-no-idle-r2 status=done' head d56a72cb50
- 2026-08-18T03:58:29Z reap removed

That head existed ONLY on the local branch: grok cannot push (lane fact), so a done task always awaits driver finalize (push+PR). The reaper removed the worktree in that window; the deliverable survived only because branch refs live in the shared object store — the driver recovered by pushing the surviving ref (PR #6995). A later ref prune would have destroyed it.

Fix at the right layer: the settled-dispatch reap path must require the worktree HEAD to be reachable from some remote ref (like the PR path requires MERGED at exact head), or defer to needs_finalize semantics for push-incapable lanes. Contrast: the PR-merged path already refuses unless PR MERGED at exact head. Stream: infra-harness (#6943).

krisztiankoos:
## Summary
Fixes #6996

When reaping settled dispatch worktrees, verify that the worktree HEAD commit is reachable from at least one remote ref (`origin/*`) before deletion:
- If HEAD is not reachable from remote (e.g. agent produced commits locally but driver has not yet pushed), skip the worktree with reason `unpushed_head` and record the skip in `batch_state/worktree-reaper/journal.jsonl`.
- Terminal tasks with 0 commits ahead of base (e.g. `dry_run`, `cancelled`, or no-op) are reachable from `origin/main` and continue to reap cleanly.
- If HEAD was pushed to `origin/<branch>`, it is reachable and reaps cleanly.

## Verification
- Unit & integration tests in `tests/orchestration/test_reap_worktrees.py` and `tests/test_fleet_post_task_reap.py` (91 passed).
- Mutation-proof test: verified that removing the `_is_head_reachable_from_remote` check causes `test_settled_dispatch_with_unpushed_local_commit_is_skipped_unpushed_head` to fail with `AssertionError: assert 'removed' == 'skipped'`.
- Clean under `ruff check`.

X-Agent: agy

### DIFF
--- a/scripts/fleet/post_task_reap.py
+++ b/scripts/fleet/post_task_reap.py
@@ -541,6 +541,14 @@ def _reap_terminal_without_pr(
             "error": None,
         }
 
+    if not reap_worktrees._is_head_reachable_from_remote(bound_path):
+        return {
+            "path": str(bound_path),
+            "action": "skipped",
+            "reason": "unpushed_head",
+            "error": None,
+        }
+
     if apply:
         live_cwds = reap_worktrees._live_cwd_paths(repo_root)
         if live_cwds is None:
--- a/scripts/orchestration/reap_worktrees.py
+++ b/scripts/orchestration/reap_worktrees.py
@@ -563,7 +563,57 @@ def _is_ancestor_of_origin_main(path: Path) -> bool:
     return proc.returncode == 0
 
 
-_TERMINAL_DISPATCH_STATUSES = frozenset({"done", "failed", "no_deliverable"})
+def _is_head_reachable_from_remote(path: Path, head: str | None = None) -> bool:
+    """Return True if ``head`` (or HEAD) is reachable from at least one origin/* ref."""
+    target = head or "HEAD"
+    proc = _run(
+        [
+            "git",
+            "for-each-ref",
+            "--format=%(refname)",
+            "--contains",
+            target,
+            "refs/remotes/origin/",
+        ],
+        cwd=path,
+    )
+    if proc.returncode != 0:
+        return False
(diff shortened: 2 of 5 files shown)


krisztiankoos:
**Cross-family review of record (Claude seat, inline vs author agy/Google): VERDICT = PASS.** `_is_head_reachable_from_remote` fails closed (git error → skip, never reap); guard sits BEFORE apply in both reap paths; terminal-status set now covers the full delegate vocabulary; zero-commit worktrees still reap correctly (their HEAD is reachable from origin/main by construction). Mutation-check quoted in the task result (guard reverted → skipped≠removed failure; restored → green); 91 tests, ruff clean. Arming into the merge queue once #6989's canary lands (queue requires merge_group CI that only #6989 puts on main).

krisztiankoos:
**Delta re-stamp (Claude seat, same reviewer): PASS.** The red consumer test was a synthetic unit test of _qualifying_reason mocking internals per the OLD contract (terminal → immediate reap, reachability never asked). Updated to the new contract with BOTH directions asserted: _is_head_reachable_from_remote=True → reap; =False → no reap (new negative case). 104+26 tests green across reap/cleanup/leak-sweep surface, ruff clean. Queueing.

PR #7037, opened 2026-08-18
fix(mcp): 405 GET/DELETE on stateless sources /mcp

### LINKED ISSUE #7036: [sources][mcp] GET /mcp hangs empty SSE; Cursor marks sources disabled
Cursor shows the project `sources` MCP as disabled. The local server is healthy (`GET /health` 200, `POST /mcp` initialize returns capabilities immediately), and `.cursor/mcp.json` already registers `http://127.0.0.1:8766/mcp`.

Root cause: the sources HTTP app runs with `stateless_http=True` and `json_response=True`. The Python MCP SDK's legacy 2025-06-18 path still accepts `GET /mcp` and opens an empty SSE stream that never emits (modelcontextprotocol/python-sdk#2474). Cursor's Streamable HTTP client waits on that GET as part of connect, then times out after 60s (`Timeout waiting for EverythingProvider with command 'mcp.createClient'`).

The Streamable HTTP spec says a server that does not offer a standalone SSE stream MUST return `405 Method Not Allowed` for GET (with `Allow: POST`). That is what Cursor needs in order to fall back to POST-only.

Fix: reject `GET`/`DELETE` on `/mcp` with 405 before the SDK transport, keep POST initialize/tools/call and legacy `/sse` unchanged, add a regression test that GET `/mcp` returns 405 within 1s instead of hanging.

krisztiankoos:
## Summary
- Cursor marked the project `sources` MCP disabled because `GET /mcp` opened an empty SSE stream and the client timed out after 60s.
- The sources HTTP app is stateless JSON (`stateless_http=True`, `json_response=True`). Per the Streamable HTTP spec, that endpoint must return `405 Method Not Allowed` with `Allow: POST` instead of hanging.
- Wrap the ASGI app so GET/DELETE on `/mcp` 405 immediately. POST initialize/tools and legacy `/sse` are unchanged.

## Test plan
- [x] `tests/test_mcp_sources_streamable_http.py` — 6 passed (GET/DELETE 405 within 1s; initialize, tools/list, tools/call, legacy `/sse` still work)
- [ ] After merge: `./services.sh restart sources`, then reload the `sources` MCP in Cursor and confirm it connects

## Related Issues
Fixes #7036

### DIFF
--- a/.mcp/servers/sources/server.py
+++ b/.mcp/servers/sources/server.py
@@ -2445,6 +2445,38 @@ async def main_stdio():
         await server.run(read_stream, write_stream, server.create_initialization_options())
 
 
+class _RejectUnsupportedStatelessMcpMethods:
+    """ASGI wrapper: 405 GET/DELETE on ``/mcp`` in stateless JSON mode.
+
+    The Python MCP SDK's 2025-06-18 Streamable HTTP path still opens an empty
+    SSE stream on GET when ``stateless_http=True`` (python-sdk#2474). Cursor
+    waits on that stream during ``mcp.createClient`` and times out, marking
+    the sources server disabled. The Streamable HTTP spec says a server that
+    does not offer a standalone SSE stream MUST return 405 Method Not Allowed
+    (with ``Allow: POST``). That is the correct POST-only signal for this app.
+    """
+
+    _MCP_PATHS = frozenset({"/mcp", "/mcp/"})
+    _REJECTED_METHODS = frozenset({"GET", "DELETE"})
+
+    def __init__(self, app: Any) -> None:
+        self.app = app
+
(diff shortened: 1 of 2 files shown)


krisztiankoos:
**Cross-family CF (AGY / Gemini) at exact head `7e3c09c2ad28b358e39b50f4439987f62665571b`.**

**VERDICT: APPROVE**

- GET/DELETE on stateless `/mcp` return 405 with `Allow: POST`; POST initialize/tools preserved
- Tests hermetic (ASGI/Starlette in-process; no network)
- No secrets/token leakage in the two-file diff

Branch CI Gate SUCCESS. Arming merge queue.

X-Agent: grok-atlas

Issue #7028, opened 2026-08-18
[sources][mcp] heritage/ESUM rank inner hits; dict search chops «» at 500 chars


krisztiankoos:
stream:infra

Local Sources MCP on `:8766`, public checkout `df0aeb81` (main). Dictionary/heritage FTS ranks inner-word hits over the headword, and `handle_dict_search` hard-truncates at 500 chars, chopping quotes.

## Repro

1. `search_definitions {query:"книга", limit:1}`
   - Expected: balanced «» / full sense 1.
   - Actual: cut mid-quote `«Що записано в кни` / `«Що записан` — 6 « vs 4 ».
   Same chop on `search_style_guide("Приймати участь")` (`…закла`).

2. `search_heritage {query:"книга", limit:2, include_live_slovnyk:false}` (~24 ms)
   - Expected: headword `книга` first.
   - Actual: Evidence 1 = ЕСУМ **`псалтир`** (score 97) because the body contains «книга псалмів». Grinchenko exact `книга` is #2 (96).

3. `search_esum("книга")` → lemmas **`рацбєя`**, **`зодія`** (“боргова книга” / “книга для ворожіння”), not `книга`. Snippets are OCR/mojibake (`2аНаг`, `родхтійрюу`).

4. Documented style-guide example **«На протязі»** (`sources_db.search_style_guide` docstring) returns no rows for `На протязі` / `протязі` / `протягом` / `слідуючий`. `Приймати участь` works. Incomplete coverage is fine; a cited example returning empty is not.

Note: MCP `include_live_slovnyk` defaults **true**; `sources_db.search_heritage` defaults **false**. First heritage call ~730 ms vs ~24 ms with `false`.

## Expected

Headword-first ranking. Quotes close. Documented style-guide examples resolve. Truncation does not split «».

## Actual

Inner-word FTS wins. Hard `[:500]` chop. Cited style-guide headword missing.

Related closed live-site ticket: #6736 (atlas lexicon quotes). Same class of chopped quotes, different surface (Sources MCP `handle_dict_search`).

`ы` in ЕСУМ/Грінченко snippets is source-faithful (citing Черных / «птица»), not tool-invented leakage.

QA 2026-08-18 ~19:40 PT. No PRs.

krisztiankoos:
[...]

PR #7040, opened 2026-08-18
fix(sources): headword-first heritage/ESUM ranking (#7028)

krisztiankoos:
Refs #7028. Headword ranks above inner-body FTS; quote-safe clip helper. X-Agent: agy/sources-7028-rank. Do not merge from this closeout.

### DIFF
--- a/scripts/wiki/sources_db.py
+++ b/scripts/wiki/sources_db.py
@@ -2424,11 +2424,17 @@ def search_esum(
             query.replace("\u0301", "").strip(),
         ]))
         meta_placeholders = ",".join("?" for _ in query_variants)
-        meta_params: list[object] = list(query_variants)
+        meta_params: list[object] = [
+            *query_variants,
+            clean_query,
+            f"{clean_query}%",
+            f"{clean_query}%",
+        ]
         meta_vol_filter = ""
         if volume is not None:
             meta_vol_filter = " AND vol = ?"
             meta_params.append(volume)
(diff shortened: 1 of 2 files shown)


krisztiankoos:
**Cross-family CF (Kimi / Moonshot) at exact head `09272caa1f9ef918ab18062a03a11d04f9111637`.**

**VERDICT: CHANGES_REQUESTED — not merging.**

Blocking: `scripts/wiki/sources_db.py:2427-2444` `search_esum` binds `volume` in the wrong placeholder order. With `volume=2`, the meta/headword path returns 0 rows (reproduced: stressed `кни́га` found without volume, dropped with `volume=2`). `.mcp/servers/sources/server.py` passes `args["volume"]` through. No test covers `search_esum(..., volume=...)`.

Non-blocking: empty lemma/headword returns tier 0 instead of 3.

Headword-first (no volume) and quote-safe clip confirmed. Tests hermetic.

Fix the param order + add a volume-filtered test. Do not merge this head.

X-Agent: grok-atlas

krisztiankoos:
**Cross-family CF re-probe (Kimi / Moonshot) at exact head `8d6ada6b4f4df96bebda092f88c83527c71a70cf`.**

**VERDICT: APPROVE**

`search_esum` volume bind order fixed. `книга` + `volume=2` returns vol-2 rows including `книгарня`. Regression test fails on the old head. 40 `test_sources_db` passed.

X-Agent: grok-infra

Issue #4700, opened 2026-07-07
[4387][EPIC] Practice & Atlas UX quality wave — post-launch bugs, gaps, and high-value improvements


krisztiankoos:
Child epic of #4387. Source: full-session audit 2026-07-07 (browser-verified on prod + dev; every item has evidence in its issue). Drive order below = my recommended sequence; checkboxes track merge-state.

## Wave 1 — verified bugs (small, ship first)
- [x] #4690 Feedback dwell: wrong answers auto-advance in 650ms — cited corrections unreadable (pedagogy-critical)
- [x] #4691 Atlas calque-note EN leakage + practice↔atlas cross-links (two ACs, one issue)
- [x] #4694 Island hydration pre-mount failure → permanently blank page (needs non-React fallback)

## Wave 2 — quick wins (well-scoped, high value/cost)
- [x] #4693 Progressive shard loading (8.9MB B1 payload; orphaned from auto-closed #4651)
- [x] #4699 Curated per-frame distractors → recover 4 starved heritage pairs (несе надо→треба)
- [x] #4697 Weak-area focus chips from the analytics we already log (§6b follow-on)

## Wave 3 — design + curation seats (need my adjudication or a spec amendment)
- [ ] #4695 Matching mode SRS integrity (design first: rate all pairs vs smaller board)
- [ ] #4698 A2 synonym curated exceptions (180 verified both-leg-A2 pairs; §9.3 amendment)
- [ ] #4506 Paronym frame authoring → unblock the fail-closed mode (6 adjudicated pairs waiting)

## Wave 4 — bigger arcs (existing issues, tracked here for the full picture)
- [ ] #4385 Atlas entry-type templates + search/redirects/counts (SSG plan PR-2/3/4 — fixes the СУМ-20 wall-of-text and expression rendering)
- [ ] #3797 Cloze sentence review throughput — A2+ cloze is empty (22 A1 items only)
- [ ] #4220 Grow judgment+publish (3,008 auto_merge candidates) — feeds heritage natives (#4505 debt), synonym legs, atlas thin pages

## User-gated (do NOT start without explicit go)
- [ ] #4696 Pronunciation audio — TTS vs recorded voice decision, then pipeline

## Out of scope here
Manifest-freshness red on main (other lane's arc, TRACK-UPDATE posted); corpus intake waves (#4593 family).

[...]

PR #7046, opened 2026-08-19
fix(practice): contrast fixes and honest empty state on Practice Hub

krisztiankoos:
## What

Three concrete UI bugs on the words-of-the-day Practice Hub (`/words-of-the-day/practice/`). Pure frontend — no cloze text, lexicon JSON, or calque/VESUM logic touched.

## Bug 1 — "Due" status text fails WCAG AA in the light theme

- `site/src/pages/words-of-the-day/practice.astro` (`.daily-deck-row.state-due .row-status`) colored the 0.75rem "НА ПОВТОРЕННЯ / DUE" status with `--lu-yellow-dark` (#E6B800) on `--lu-surface` (#FFFFFF) — **1.9:1 contrast**. The repo itself documents this failure (`course.css` rule-box note: "#E6B800 fails WCAG AA on --lu-yellow-light").
- **Repro:** light theme → open the "Показати слова" word list with any due row → the amber status text is barely readable.
- **Fix:** new `--lu-yellow-text` token (`course.css`): #8A6D00 in light (4.9:1), #FFE58A in dark (~14:1, matching the existing dark fill).

## Bug 2 — White-on-light-blue selected controls in the dark theme

- Four filled selected/active states used `--lu-on-primary`, which stays **white in both themes** (`custom.css:174`), on `--lu-primary` (#5B9BD5 in dark) — **2.9:1**:
  - `.lexicon-practice-modes/levels-row button.active` (practice.astro:144)
  - `.lexicon-practice-stage [data-selected="true"]` (match tiles, :975)
  - `.stress-vowel.selected` (:2093)
  - `.zno-practice-next` (:2318)
- **Repro:** dark theme → start a session → pick a match tile / stress vowel, or look at the active mode/level pill → white text on light blue.
- **Fix:** switch all four to `--lu-on-primary-fill` — the token the #6814 fix established for exactly this case (white on azure in light, dark on the light-blue dark primary, 6.1:1).

## Bug 3 — Empty daily deck renders "1 з 0" and navigates to index −1

- `PracticeDailyDeck.tsx` rendered the position counter unconditionally: with `snapshot.items.length === 0` the header read **"1 з 0"**, and clicking ‹ wrapped `previewIndex` to `total - 1` = **−1** ("0 з 0"). The card front showed a bare "—".
- **Repro:** any state where the daily pool is empty (e.g. a custom deck with no eligible words at the current level) — `LexiconPractice.tsx:4071` mounts the deck whenever the snapshot is non-null, with no item-count guard.
- **Fix:** counter hidden at 0 items, nav buttons disabled, handlers no-op, and the card front shows the existing `practice.noCards` copy ("Поки що немає карток для практики.").

## Screenshots-in-words

- Before (light): word-list row "✦ 1 мама … ДО ПОВТОРЕННЯ" with the status in pale gold #E6B800 on white — washes out at a glance. After: same row, status in deep amber #8A6D00, clearly readable.
- Before (dark): selected stress vowel / "Далі" ZNO button = white glyph on #5B9BD5, low contrast. After: near-black #1A1A1A text on the same blue fill.
- Before (empty deck): header "Слова дня — 1 з 0", clickable ‹ › around an em-dash card; clicking ‹ flips the counter to "0 з 0". After: no counter, greyed-out disabled ‹ ›, card reads "Поки що немає карток для практики."

## Verification

- `vitest run tests/unit/LexiconPractice.test.tsx` — **178 passed**, including a new regression test: empty deck → no "1 з 0", nav disabled, `practice.noCards` copy shown, clicking prev stays empty.
- `k3-chrome-locale`, `practice-locale-purity`, `atlas-hero-contrast` — all pass (k3-chrome-locale asserts the `--lu-on-primary-fill` precedent this PR extends).
- `tsc --noEmit` — only the 16 pre-existing errors in untouched files; none in the changed files.

Refs #4700 (UX wave) — does not close it.

X-Agent: kimi/practice-frontend

### DIFF
--- a/docs/lesson-schema.yaml
+++ b/docs/lesson-schema.yaml
@@ -2,7 +2,7 @@ schema_version: "1.0"
 generator_version: "1.0.0"
 generated_from:
   components_dir: site/src/components/{*.tsx, *.astro}
-  components_sha256: 17636cd8d4f6115ac1b314ca8c00bbb1f7183b544e44cb13f165d6e3158b631d
+  components_sha256: 3d13b1df624d458a7671a968d9190ca299207d63925e77c6219fc15fcaab4e3b
   config_tables_sha256: c9621b1f9b95e7fe7d7400989839ef3a1ecf338f8cec8989ae9c53bde8c9b287
   lesson_contract_sha256: 917ee4abd1a74c5df17f9d19e743f23c376f43d8f1b9a68e12fb30ad6b535024
 components:
--- a/site/src/components/PracticeDailyDeck.tsx
+++ b/site/src/components/PracticeDailyDeck.tsx
@@ -68,11 +68,13 @@ export default function PracticeDailyDeck({
   const currentLexeme = currentLemmaId ? lexemes.get(currentLemmaId) ?? null : null;
 
   const handlePrevious = () => {
+    if (total === 0) return;
     setFlipped(false);
     setPreviewIndex((index) => (index > 0 ? index - 1 : total - 1));
   };
 
   const handleNext = () => {
+    if (total === 0) return;
     setFlipped(false);
     setPreviewIndex((index) => (index < total - 1 ? index + 1 : 0));
   };
@@ -108,12 +110,14 @@ export default function PracticeDailyDeck({
             <ChromeText k="practice.wordsTitle" />
           )}
         </h2>
-        <span className="daily-deck-position" aria-live="polite">
-          <ChromeDual
-            uk={`${previewIndex + 1} з ${total}`}
-            en={`${previewIndex + 1} / ${total}`}
-          />
-        </span>
+        {total > 0 ? (
+          <span className="daily-deck-position" aria-live="polite">
+            <ChromeDual
+              uk={`${previewIndex + 1} з ${total}`}
+              en={`${previewIndex + 1} / ${total}`}
+            />
+          </span>
+        ) : null}
         {onReRoll ? (
           <button
             type="button"
@@ -135,6 +139,7 @@ export default function PracticeDailyDeck({
           type="button"
           className="daily-deck-nav"
           aria-label={chromeString(chromeLocale, 'label.previous')}
+          disabled={total === 0}
           onClick={handlePrevious}
         >
           ‹
@@ -162,7 +167,9 @@ export default function PracticeDailyDeck({
                   {frontSubtitle && <span className="flashcard-subtitle">{frontSubtitle}</span>}
                 </>
               ) : (
-                <span className="flashcard-word">—</span>
+                <span className="flashcard-subtitle" data-testid="practice-daily-empty">
+                  <ChromeText k="practice.noCards" />
+                </span>
               )}
             </div>
             <div className="flashcard-back">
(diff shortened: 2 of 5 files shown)


krisztiankoos (commented):
test

krisztiankoos (commented):
CF review by Claude (cross-family, not self-review — author is Kimi/Moonshot). Reviewed at head `6346d8f65075f022b58be41f4412addd82a1e122`. Scope: Practice Hub UI (`PracticeDailyDeck.tsx`, `practice.astro`, `course.css`, schema hash, unit test). Note: this agent cannot post a formal GitHub "Approve"/"Request changes" review (`AGENT_NO_MERGE=1`, incident #1403) — verdict below is the review of record; a human/self-merge step still applies the GitHub state.

**Verdict: APPROVE**

**Empty-deck fix (`PracticeDailyDeck.tsx`)** — correct root-cause fix. Previously `handlePrevious`/`handleNext` could wrap `previewIndex` to `-1` when `total === 0` (the old empty face was a bare `—`, the position pill read "1 з 0"). Now: nav handlers no-op on `total === 0`, nav buttons get `disabled`, the position pill is hidden entirely instead of rendering a nonsensical count, and the empty card face uses the existing `practice.noCards` chrome key (verified present in both locales in `chrome.ts`, and already reused elsewhere in `LexiconPractice.tsx`). The new unit test covers the "no 1 з 0", disabled-nav, and no-wrap-to-negative cases and matches the actual UK copy in `chrome.ts`.

**Contrast fixes (`practice.astro` + `course.css`)** — checked the WCAG math myself rather than trusting the inline comments:
- `--lu-on-primary` (white, fixed) on dark-theme `--lu-primary` (`#5B9BD5`) ≈ 2.96:1 — genuinely fails AA. Swap to the existing `--lu-on-primary-fill` token (already defined in both theme blocks, `#FFFFFF` light / `#1A1A1A` dark) fixes it to ≈5.9:1 dark, ≈6.9:1 light. Correct reuse of an existing token, not a new one.
- `--lu-yellow-dark` (`#E6B800`) as text on light-theme `--lu-surface` ≈ 1.87:1 — fails badly. New `--lu-yellow-text` token: ≈4.9:1 light theme, ≈13.8:1 dark theme. Both pass AA, stays in-family with the flag palette, and the dark-theme value tracks the existing yellow tint rather than inventing a new hue.

**Schema hash** — `components_sha256` bump is CI-validated (`Contracts (schema, MDX, atlas, BIO)` green), not hand-edited noise.

CI is green across the board (Frontend build+vitest, Playwright E2E, Contracts, Ruff).

Two non-blocking observations, not requesting changes:
1. The flashcard **back** face still falls through to a bare `—` in the empty-deck state (only the front got the `practice.noCards` swap). Flipping an empty card via the whole-card `onClick` isn't gated by `total === 0`, so a tap still reveals the old dash. Low severity, not user-facing on first paint.
2. `total` is `snapshot.items.length` while the rendered card comes from `orderedRows` (`pendingDue + pendingNew + done`). If those ever diverge (nonzero `total`, empty `orderedRows`), the empty-card message would show next to a nonzero position count. Pre-existing, out of scope for this PR.

Issue #7048, opened 2026-08-19
[infra] Step 1: VPS loopback services + SSH tunnel (public + private services.sh)


krisztiankoos:
## Parent
#6943 (infra area)

## Public-safe scope
Loopback **code** in this public repo so services cannot bind a public interface. Operational VPS topology, SSH tunnel inventory, host class, and agent-access runbook live in the **private** repo: learn-ukrainian/learn-ukrainian-infra-private#516.

Do **not** add hostnames, IPs, provider SKUs, disk sizes, or tunnel diagrams to this issue or to public docs.

## Acceptance (code only)
- [ ] Public `services.sh` `api` binds loopback (`127.0.0.1:8765`); `SVC_MATCH[api]` matches the new argv (today it still matches `0.0.0.0`).
- [ ] `sources` bind is explicit loopback. `work` / `astro` stay loopback.
- [ ] `package.json` `api` / `api:bg` and `docs/best-practices/local-api-server.md` stop prescribing `0.0.0.0`.
- [ ] `GET /api/health` includes host identity + git SHA (so a leftover local process is distinguishable). No IPs/secrets in the payload.
- [ ] In-repo systemd **templates** that invoke existing `services.sh` semantics; no IPs/hostnames in git.
- [ ] Tests for bind/match/health identity.

## Non-goals
- VPS provision
- Public runbook of topology (private #516)
- Fleet worktrees on a remote host
- Grok Bot orchestrator
- Closing product issues

Related: closed-unmerged PR #7047.

X-Agent: grok-atlas

krisztiankoos:
Runbook PR: #7047. Implementation starting (loopback bind + health identity + systemd templates + onboarding). VPS provision stays operator. X-Agent: grok-atlas

krisztiankoos:
Runbook and host/tunnel/agent-access plan moved to private issue learn-ukrainian/learn-ukrainian-infra-private#516 after operator correction: confidential ops docs do not belong in the public tree. This issue is loopback code only. X-Agent: grok-atlas

krisztiankoos:
Loopback code PR: #7049 (`f706c478ad`). AGY CF in flight. Ops runbook stays private #517. X-Agent: grok-atlas

krisztiankoos:
[...]

PR #7052, opened 2026-08-19
infra: opaque Monitor occupancy schema (#7050)

### LINKED ISSUE #7050: [infra] Occupancy fields on Monitor load/health (code only)
## Parent
#6943

## Public-safe scope
Code-only Monitor JSON so occupancy can be `{kind, agent, task_id, epic}` plus an **opaque** `host_id` (from env / instance id, not an address). Operational host map, SSH, and pilot sequencing live in the **private** repo only (learn-ukrainian-infra-private#518, #519).

Do **not** add IPs, SSH hostnames, provider SKUs, disk sizes, role-to-machine rosters, work roots, or tunnel diagrams to this issue or to public docs.

## Acceptance (code only)
- [ ] One occupancy schema on Monitor (extend an existing load endpoint **or** one sibling route — not two boards).
- [ ] Occupants are `{kind, agent, task_id, epic}` with `kind` in `driver | worker | job | service`.
- [ ] `host_id` is opaque. Tests and MONITOR-API examples use placeholders (`<host-id>`), never real names or addresses.
- [ ] Unreachable probe → `unavailable` with no SSH/error text. No sync remote probe on the request path.
- [ ] Fixtures contain no IPs.

## Non-goals
- Host inventory / topology
- Provisioning
- Orchestrator migration
- Public runbook of where jobs run

X-Agent: cursor/vps-occupancy-intake

krisztiankoos:
## Summary

Adds `GET /api/occupancy` so drivers can see who is on which host without SSH aliases, addresses, or a second probe board. Reuses the existing atlas-jobs load cache.

## Changes

- New occupancy schema `monitor-occupancy.v1` with opaque `host_id` keys from `MONITOR_OCCUPANCY_HOST_IDS`
- Occupants are `{kind, agent, task_id, epic}` (`driver | worker | job | service`)
- Unreachable hosts return `unavailable` / `unreachable` with no SSH text and no load metrics
- MONITOR-API examples use placeholders only (`host-job`, `host-teacher`)

## Testing

- [x] `.venv/bin/python -m pytest tests/api/test_occupancy.py tests/test_monitor_route_contracts.py tests/api/test_atlas_jobs_router.py tests/test_api_endpoints.py` — 97 passed
- [x] `ruff check` on occupancy files
- [ ] Modules pass audit (`scripts/audit_module.py`) — N/A (Monitor API)
- [ ] Website builds without errors (`cd site && npm run build`) — N/A
- [ ] Ukrainian text reviewed for naturalness — N/A

## Related Issues

Closes #7050


krisztiankoos:
## Verdict: REQUEST-CHANGES

Independent cross-family review (Claude vs Cursor/Grok authorship). Read-only: nothing pushed, nothing merged.

The core security property the PR is built on — no SSH on the request path — **holds and is correctly implemented**. `occupancy_payload` reads only `load_mod._get_host_load_entry`, which serves from `_HOST_LOAD_CACHE` and hands refresh to `asyncio.create_task` (`scripts/api/atlas_jobs_router.py:137-192`). Unreachable hosts get a hardcoded `error: "unreachable"` with metrics dropped (`scripts/api/occupancy.py:148-150`). No second probe board. `Cache-Control: no-store`. 13/13 tests pass locally against head `68e4a8e`. Commit carries the `X-Agent` trailer.

What blocks is narrower: the opacity invariant is enforced on JSON *keys* but not on JSON *values*, the issue's own fixture rule is violated, and the tests that look like they guard opacity do not.

---

## Blocking

### B1 — Real canonical host names in the public test fixture; the deploy mapping is now public

`tests/api/test_occupancy.py:19`

```python
_PLACEHOLDER_MAP = "atlas-runner=host-job,hramatka=host-teacher"
```

Issue #7050 acceptance says verbatim: *"`host_id` is opaque. Tests and MONITOR-API examples use placeholders (`<host-id>`), never real names or addresses."* The new test file contains the real canonical names 13 times (lines 18, 19, 26, 42, 43, 119, 148, 195, 197, 213, 214, 218, 219). The PR body's claim "Tests/docs use host-job / host-teacher only" is not accurate.

The aliases themselves are already public in ~30 repo files, so no new *name* is disclosed. What is new is the **pairing**: `host-job → atlas-runner`, `host-teacher → hramatka` is now published verbatim, and that pairing is exactly what the opaque scheme exists to hide. Anyone reading the public test can de-anonymize a production response built on the same env value.

This is avoidable, not inherent. `_canonical_host` passes unknown names through unchanged (verified: `parse_host_id_map("typo-not-a-real-host=host-x")` → `{'typo-not-a-real-host': 'host-x'}`), so the tests can warm the cache under fictional canonical keys (`example-a=host-job`) and exercise every path without a real name. The one test that legitimately needs a real alias is the `_CANONICAL_ALIASES` rejection test, which can reference the constant instead of literals.

### B2 — `_safe_field` does not enforce the invariant it exists for

`scripts/api/occupancy.py:59-67`

The occupant sanitizer blocks paths and IPv4, but not canonical aliases, not IPv6, not FQDN-shaped hostnames. `_opaque_host_id` (line 51) applies both the `_CANONICAL_ALIASES` denylist and the IP check; `_safe_field` applies neither. Verified against head:

```
'atlas-runner'              -> 'atlas-runner'
'hramatka'                  -> 'hramatka'
'vps'                       -> 'vps'
'2001:db8::1'               -> '2001:db8::1'
'vps.example-provider.net'  -> 'vps.example-provider.net'
'10.0.0.1'                  -> None      # only this class is caught
'/home/user/secret'         -> None
```

So `_occupant(kind="job", agent="hramatka", task_id="j1", epic="atlas-runner")` emits `{"kind":"job","agent":"hramatka","task_id":"j1","epic":"atlas-runner"}` — canonical aliases in the response body of the endpoint whose docstring says "Canonical SSH aliases never appear as JSON keys." True for keys; not true for the payload. This is reachable through operator-chosen job ids (`_SAFE_TOKEN` matches `atlas-runner-reenrich-3`), `row["agent"]`, and `row["epic"]`.

Fix: route `_safe_field` through the same alias + address rejection as `_opaque_host_id` (substring match, case-folded, plus an IPv6 check), and derive `_CANONICAL_ALIASES` from `atlas_job.ALLOWED_HOSTS`/`HRAMATKA_ALIASES` instead of the hand-copied literal at line 28 — the static copy drifts the moment a host is added, silently disabling the denylist for the new one.

### B3 — The tests do not lock the sanitization contract, and one assertion is vacuous

Six tests cover empty map, happy shape, unavailable, registry occupants, unknown `host_id`, and stale reuse. None of them exercise `_safe_field`, `_opaque_host_id`, or `parse_host_id_map` with hostile input. The two security-critical validators have zero direct coverage; every fixture is benign, so the `assert alias not in text` checks pass for reasons unrelated to the sanitizer working. B2 is invisible to this suite.

Worse, `tests/api/test_occupancy.py:197`:

```python
assert "atlas-runner" not in resp.text or resp.json()["detail"] == "unknown host_id"
assert resp.json()["detail"] == "unknown host_id"
```

Line 197 can never fail while line 198 passes — the `or` makes the echo-check unconditionally true. It reads like a guard against the 400 response echoing the caller-supplied alias, and asserts nothing. The strict form (`assert "atlas-runner" not in resp.text`) passes today and should replace it.

Minimum additions to call the contract locked: hostile occupant fields (alias, IPv6, FQDN, path, oversize, empty `task_id`); `parse_host_id_map` rejecting `atlas-runner=vps` and `atlas-runner=10.0.0.1` (both already work — lock them); duplicate-opaque and malformed entries; and `?fresh=true` not blocking.

---

## Non-blocking

1. **Three of four occupant kinds are unreachable.** Both producers hardcode `kind="job"` (`occupancy.py:101,122`). `driver`, `worker`, and `service` can never be emitted, so the endpoint cannot answer "which driver is on this host" — the motivating use case in the PR summary. Acceptable as v1 if the docs say so; right now `docs/MONITOR-API.md:238` advertises all four without qualification. Please add a follow-up issue and a one-line doc caveat.

2. **Occupancy silently under-reports, which is the dangerous direction for a dispatch board.** `_occupants_from_job_unit` (line 116-123) takes `job_unit.job_id` — the *first* active unit — and discards `active_count`, so a host with three active units reports one occupant. And `except Exception: rows = []` (line 87-90) turns a registry read failure into an empty occupant list on a `status: fresh` host. A driver reading `{"status":"fresh","occupants":[]}` concludes "idle" and dispatches onto a saturated host. Surface `active_count`, and distinguish "registry unreadable" from "nobody home".

3. **No validation of the canonical side of the env map,** unlike the sibling `/api/atlas-jobs/load`, which rejects hosts outside `_canonical_allowed_hosts()`. Verified behaviours worth hardening: `"=host-x"` silently becomes `{'atlas-runner': 'host-x'}` (empty string falls through `_canonical_host`'s `host or "atlas-runner"` default at `atlas_job.py:516`); `"atlas-runner=host-job,hramatka=host-job"` is accepted and then collapses to a single host in the `reverse` dict at `occupancy.py:166`, silently dropping one host from the board; unknown canonical names are accepted and produce a permanently-`unavailable` phantom host plus a background probe to a nonexistent alias.

4. **Empty map plus `host_id` returns 200 with `{}` instead of 400.** With a map, an unknown id is a 400; without a map, every id is a silent success. A client cannot distinguish "not configured" from "no such host" from "no hosts". Consider a `configured: false` flag or consistent 400. Untested either way.

5. **Doc wording overstates the SSH property.** "no SSH on the request path" (`docs/MONITOR-API.md:236`) reads as "this endpoint never causes SSH". A cache miss, a stale entry, or `?fresh=true` all schedule `adapter.host_load()`, which shells out to `ssh` (`atlas_job.py:419`). Accurate phrasing: never *blocks* on SSH; refresh happens off-path.

6. `_occupants_from_job_unit(canonical, load_entry)` takes `canonical` only to `del` it on line 117. Drop the parameter.

7. `_canonical_host(None)` returns `"atlas-runner"`, so a registry row with a missing `host` is attributed to that host by the filter at line 94. Compare against a sentinel instead.

8. `_shape_host` copies `mem` and `disk` verbatim (line 151-153). Today those are numeric-only, so nothing leaks — but the allowlist is shallow, and an adapter change that adds a mount path inside `disk` would pass straight through. Allowlist the inner keys too.

9. The new docs section landed under "Health & Config" (`MONITOR-API.md:234`), far from the `/api/atlas-jobs` section at line 3177 where its sibling `load` endpoint is documented.

---

## Deploy question: `MONITOR_OCCUPANCY_HOST_IDS=atlas-runner=host-job,hramatka=host-teacher`

**Safe to run, but do not use those two ids.** Mechanically the endpoint is sound with that env: both canonical names resolve, the response keys are `host-job`/`host-teacher`, no IPs or SSH text appear, unavailable hosts carry no metrics, the service is loopback-only per #7048, and nothing SSHes on the request path.

The problem is that this exact string is `_PLACEHOLDER_MAP` in the public test file (B1), so the opaque ids are pseudonyms with a published key, not secrets. If opacity is meant to be more than cosmetic, deploy ids that do not appear anywhere in the public repo and keep the fixture values fictional.

Two expectations to set before drivers rely on it: it will list at most one job per host from the probe and zero drivers (non-blocking 1 and 2), so an empty `occupants` array does not currently mean the host is free.

---

Happy to re-review promptly once B1–B3 are addressed; they are all small, local changes.

krisztiankoos:
## Verdict: REQUEST-CHANGES

Re-review of head `35922ce1b4` (independent cross-family, Claude vs Cursor/Grok). Read-only: nothing pushed, nothing merged.

Two of the three blocking findings are properly fixed, and the PR body claim was corrected. But the remediation commit turned CI red on a hard gate, only half-fixed B2, and introduced two silent data-loss paths that did not exist at `68e4a8e`. All of it lives in one function plus one lint fix.

### Fixed — verified at exact head

- **B1 (fixture pairing) — fixed.** `_PLACEHOLDER_MAP = "job-box=host-job,teach-box=host-teacher"` (`tests/api/test_occupancy.py:19`). No real alias is paired with an opaque id anywhere in the file. The residual `atlas-runner` references (the `_ALIAS_LEAKS` tuple and the `?host_id=atlas-runner` rejection test) are already-public names used as negative assertions, which is the correct use. Cache warming under the fictional keys works because `_canonical_host` passes unknown names through.
- **B3 (tests) — fixed.** The vacuous `or`-assertion at old line 197 is gone, replaced with a real loop over `_ALIAS_LEAKS` against the 400 body. `test_parse_host_id_map_drops_non_opaque_values` and `test_safe_field_drops_aliases_addresses_and_fqdn` are genuine unit tests on both validators, including the `_occupant(agent="hramatka", epic="atlas-runner")` → both-fields-`None` case. 31/31 pass locally (`test_occupancy.py`, `test_monitor_route_contracts.py`, `test_atlas_jobs_router.py`).
- PR body no longer claims tests use placeholders only; `X-Agent` trailer present.

---

## Blocking

### R1 — CI is red: the fix breaks the Ruff gate

The `Ruff` check on this head is `FAILURE` ([job](https://github.com/learn-ukrainian/learn-ukrainian.github.io/actions/runs/32301291282/job/96224170890)). Reproduced locally on the PR's own files:

```
SIM103 Return the negated condition directly
  --> scripts/api/occupancy.py:57:5
57 | /     if _IPV4.search(value) or _IPV6.search(value) or "." in value:
58 | |         return False
59 | |     return True
```

`ruff check scripts/api/occupancy.py tests/api/test_occupancy.py` exits 0 at `68e4a8e` and 1 at `35922ce`, so this is introduced by the remediation commit, not inherited. Repo hard gates bind — this cannot merge as-is.

### R2 — B2 is only half fixed: aliases still leak as substrings

`scripts/api/occupancy.py:64` uses exact, case-folded membership:

```python
if text.lower() in _CANONICAL_ALIASES:
    return None
```

My prior finding asked for a **substring** match and gave `atlas-runner-reenrich-3` as the concrete reachable vector. That exact value still passes through verbatim. Verified at head:

```
'atlas-runner-reenrich-3' -> 'atlas-runner-reenrich-3'
'atlas-runner-1'          -> 'atlas-runner-1'
'hramatka-drive'          -> 'hramatka-drive'
'job-on-hramatka'         -> 'job-on-hramatka'
'vps-2'                   -> 'vps-2'
'hramatka'                -> None      # exact match is caught
'ATLAS-RUNNER'            -> None      # case-folding works
```

Since `_SAFE_TOKEN` permits `-`, any hyphenated derivative of a canonical alias reaches the payload through an operator-chosen job id, `row["agent"]`, or `row["epic"]`. The new test only asserts the exact-match cases, which is why the residual survived — the test was written to the implemented fix rather than to the finding.

Security impact is genuinely low (already-public names, loopback-only endpoint), so I would not block on this alone. I am blocking because it sits in the same function as R1 and R3 and the fix is one line: `any(alias in text.lower() for alias in _CANONICAL_ALIASES)`.

### R3 — New silent occupant loss from the tightened charset

`_SAFE_TOKEN` was narrowed from `[A-Za-z0-9._:-]` to `[A-Za-z0-9_-]` and a blanket `"." in text` reject was added (`occupancy.py:26,67`). Two consequences that did not exist at `68e4a8e`:

**(a) A contract-legal job id deletes the whole occupant.** Upstream `atlas_job._SAFE_ID` is `^[A-Za-z0-9][A-Za-z0-9._-]{0,62}$` — **dots are permitted in job ids** by `require_safe_job_id`. `_safe_field` now rejects them, and `_occupant` returns `None` when `task_id` is `None`, so the occupant vanishes entirely rather than losing one field. Verified:

```
_safe_field("v2.1-reenrich")                                  -> None
_occupant(kind="job", task_id="v2.1-reenrich", epic="atlas")  -> null
```

A host running that job reports `occupants: []` with `status: fresh`. This is the failure direction I flagged last round as the dangerous one for a dispatch board: a saturated host renders as idle. A dot in a job id is not an address — please don't apply the FQDN rule to `task_id`, and when a field is unusable, drop the field rather than the occupant.

**(b) `hramatka` is also a first-class epic slug, so real epic labels are nulled.** `scripts/session_canary/grok_lane.py:72` maps `"hramatka" → "epic:4542"`, `scripts/fleet/hramatka_scope_gate.py:27` sets `HRAMATKA_EPIC = 4542`, and `hramatka.epic` is a named epic assignment in `agents_extensions/shared/agents/infra-orchestrator.md`. The host-alias denylist is applied to `epic`, so:

```
_occupant(kind="job", task_id="ok-1", epic="hramatka") -> {"epic": null, ...}
```

An already-public curriculum track name is discarded for zero security gain, because the denylist conflates a host namespace with an epic namespace. Latent while only `kind="job"` exists; it bites immediately when `kind="driver"` lands, since a curriculum-orchestrator drive on epic hramatka is exactly the occupant a driver wants to see. Scope the alias denylist to fields that can carry host identity, or exempt `epic`.

---

## Nits

1. The new `_IPV4` / `_IPV6` / `"." in value` guards inside `_opaque_host_id` (`occupancy.py:57`) are unreachable — `_OPAQUE_HOST_ID` (`^[a-z][a-z0-9-]{0,62}$`) already rejects every dot, colon and dotted quad. Verified: all three candidates fail the regex before reaching the guards. Harmless, but it is the line Ruff is flagging, so it can simply go.
2. `_IPV6 = re.compile(r":")` is a misnomer — it rejects any colon, not IPv6 specifically. Given `_SAFE_TOKEN` no longer admits colons, it is redundant in `_safe_field` too. Name it for what it does or fold it into the charset.
3. Round-1 non-blocking items 1–9 are unaddressed, which is fine for this scope. The two worth keeping on the follow-up: only `kind="job"` is ever emitted, and `job_unit.active_count` is discarded so the probe yields at most one occupant per host.

---

## Deploy answer, unchanged in substance

With B1 fixed, `MONITOR_OCCUPANCY_HOST_IDS=atlas-runner=host-job,hramatka=host-teacher` is no longer de-anonymized by the repo, so those ids are now usable — though ids absent from public git remain the stronger choice. The safety properties I verified in round 1 still hold at this head: no request-path SSH, no metrics or error text on unavailable hosts, opaque keys, `no-store`.

The caveat is now larger, not smaller: because of R3, an empty `occupants` array means even less than it did before. Do not let a driver treat this endpoint as authoritative for "host is free" until R3 and the `active_count` gap are closed.

---

Re-review turnaround will be quick — R1 through R3 are one lint fix and one function.

krisztiankoos:
## Verdict: APPROVE-WITH-NITS

Re-review of head `f0c04654623c385033cfaf2a09c84217fff5836f` (independent cross-family, Claude vs Cursor/Grok). Read-only: nothing pushed, nothing merged.

All three blockers from `35922ce` are fixed, and the role-based sanitizer is the right shape — the previous round's mistake was applying one rule to three fields with different meanings, and splitting `agent` / `task_id` / `epic` addresses that at the root rather than patching symptoms. Verified at the exact head, not from the diff.

### R1 — Ruff: fixed

`_opaque_host_id` is now a single `return bool(...)` (`scripts/api/occupancy.py:58-65`). `ruff check scripts/api/occupancy.py tests/api/test_occupancy.py` exits 0, and the CI `Ruff` job is SUCCESS at this head.

### R2 — Alias tokens: fixed

`_ALIAS_TOKEN` (`occupancy.py:31-34`) matches on non-alphanumeric boundaries with `IGNORECASE`, applied to `agent` and `task_id`. Every vector I raised is now dropped:

```
'atlas-runner-reenrich-3'  agent=None  task_id=None
'atlas-runner-1'           agent=None  task_id=None
'hramatka-drive'           agent=None  task_id=None
'job-on-hramatka'          agent=None  task_id=None
'vps-2'                    agent=None  task_id=None
'ATLAS-RUNNER-JOB'         agent=None  task_id=None   # case-folding
'atlas-runner.1'           agent=None  task_id=None   # dot delimiter
```

`_occupant(kind="job", task_id="atlas-runner-reenrich-3")` → `None`.

### R3 — Silent occupant loss: fixed

`_TASK_TOKEN` admits dots for `task_id` while `_FQDN` still blocks hostname-shaped values, so a contract-legal dotted job id survives instead of deleting the occupant:

```
task_id 'v2.1-reenrich'   -> 'v2.1-reenrich'
task_id 'job.1'           -> 'job.1'
occupant -> {"kind":"job","agent":null,"task_id":"v2.1-reenrich","epic":"atlas"}
```

And the epic slug collision is gone: `epic="hramatka"` → `"hramatka"`, occupant intact. Addresses remain blocked in **all three** roles (`10.0.0.1`, `2001:db8::1`, `box.example.com`, `host.internal`, `/home/ops/job` → `None` for agent, task_id and epic). An unrecognized `role` falls back to the strict token plus the alias check, so the default is fail-closed.

The tests lock exactly these cases: `v2.1-reenrich` kept as `task_id`, `hramatka` kept as `epic`, the three alias-token values asserted `None` in both `agent` and `task_id` roles, and the occupant-drop assertion for an alias `task_id`. 99 tests pass locally (`test_occupancy.py`, `test_monitor_route_contracts.py`, `test_atlas_jobs_router.py`, `test_api_endpoints.py`).

---

## Nits — none blocking, none need to hold the merge

1. **`epic` is now blanket-exempt from the alias check, which is wider than the collision required.** `occupancy.py:79` skips `_ALIAS_TOKEN` for the whole field, so anomalous epic values pass verbatim: `epic="atlas-runner"` → `"atlas-runner"`, `epic="vps"` → `"vps"`, `epic="atlas-runner-reenrich-3"` → `"atlas-runner-reenrich-3"`. My finding was narrower — `hramatka` is a legitimate track slug (`epic:4542`) and shouldn't be nulled — so exempting just the colliding token, or checking `epic` against known slugs, would keep the guard while fixing the collision. Note the previous test asserted `epic="atlas-runner"` → `None`; that assertion was replaced rather than kept alongside the new one, so this case lost its coverage. Severity is very low: already-public names, loopback-only endpoint, and an epic named after a host is a data-entry anomaly rather than a leak path.

2. **Token-boundary corners pass by design:** `atlas-runner2`, `hramatka2`, `vpsx` are not matched, since `_ALIAS_TOKEN` requires a delimiter. That is the trade-off I asked for over blanket substring matching (which would over-reject), so I'd leave it — just worth knowing the boundary is a delimiter, not a prefix.

3. `role: str` could be `Literal["agent", "task_id", "epic"]` to make the three call sites type-checked. Behaviour is already fail-closed on a typo, so this is cosmetic.

4. Round-1 non-blocking items remain open and are fine to carry into the follow-up that adds the other occupant kinds. The two that matter most: only `kind="job"` is ever emitted, and `job_unit.active_count` is discarded so the probe contributes at most one occupant per host. Worth a doc caveat at `docs/MONITOR-API.md:238`, which advertises all four kinds without qualification.

---

## Deploy

`MONITOR_OCCUPANCY_HOST_IDS=atlas-runner=host-job,hramatka=host-teacher` is safe to deploy after merge. Everything I checked in round 1 still holds at this head — no request-path SSH, opaque keys, no metrics or error text on unavailable hosts, `no-store` — and the sanitizer now actually enforces the invariant it advertises for occupant values. Ids absent from public git remain the marginally stronger choice, but the repo no longer publishes this pairing.

One expectation to keep setting for drivers: with only `kind="job"` implemented and `active_count` unused, an empty `occupants` array still does not prove a host is free. That is a v1 scope limit, not a defect in this PR.

Merge gate: `Ruff` is green; `Pytest fastlane`, `Contracts`, and the CodeQL Python/JS analyses were still in flight when I reviewed, so let those finish before merging. I verified the pytest substance locally in the meantime.

Nice root-cause fix on the role split — thanks for taking the sanitizer apart properly rather than adding another guard on top.

PR #6809, opened 2026-08-14
feat(atlas): promote textbook VESUM-ok leftovers (#6371)

krisztiankoos:
## Summary

Phase 1 of #6371 measures and promotes the committed textbook-intake inventory through the existing source-inventory -> VESUM -> manifest -> enrich -> DB/search build path.

- Honest denominator: 24,682 approved rows in the committed curated textbook intake inventory.
- Promoted: 7,750 VESUM-ok rows absent from the current 20,112-entry hydrated manifest.
- Enriched: all 7,750 promoted entries.
- Refreshed: the established DB-backed search, alias, browse-meta, flagged, and browse-letter artifacts (34 generated JSON files).
- No release upload and no lexicon-manifest.pointer.json change; live pointer publication remains with the driver/operator.

## Denominator and residual

The denominator is the committed textbook-intake row set. The raw textbook prose/token universe is not committed in this sparse dispatch worktree, so this PR makes no claim about that larger universe.

| Class | Count | Phase-1 disposition |
|---|---:|---|
| Textbook intake rows | 24,682 | measured |
| Already in Atlas before promotion | 16,932 | covered |
| VESUM-ok, missing before promotion | 7,750 | promoted and enriched |
| VESUM absent | 0 | no residual |
| Candidate rows needing review | 0 | no residual |
| Missing candidate rows | 0 | no residual |
| Uncovered after promotion | 0 | cleared by post-plan |

The remaining downstream residual is intentional scope: practice/deck publication is not run in this bounded Phase 1, and the release pointer remains operator-owned. Those are not intake blockers.

## Evidence preamble (#M-4)

All commands below ran from the dispatch worktree with the shared primary interpreter.

Candidate build:

~~~
/Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/python -c 'from pathlib import Path; from scripts.lexicon.curated_textbook_jsonl_repromote import build_candidates; build_candidates(Path("data/lexicon/source-inventory/oneshot/textbook-jsonl-curated-2026-07-19-bulk.yaml"), Path("/tmp/atlas-6371-textbook-candidates.json")); print("candidate_payload=/tmp/atlas-6371-textbook-candidates.json")'
candidate_payload=/tmp/atlas-6371-textbook-candidates.json
~~~

Pre-promotion locked planner:

~~~
[...]

Issue #6371, opened 2026-08-05
[4220][P3] Textbook leftover lemmas → Atlas + enrich + practice


krisztiankoos:
## Goal
Remaining textbook-curated lemmas not covered by teacher/Ohoiko passes enter Atlas the same way.

## Acceptance
- [ ] Residual textbook-intake lemmas not in Atlas = 0 (or tool-proved blockers)
- [ ] Enriched + practice + deck publish

## Parent
#4220 · #4387
## Depends on
P1 + P2

krisztiankoos:
## Driver / AGY Remeasure: Textbook Leftover Residual vs Live Pointer

Task **atlas-6371-remeasure** (measure-only): Honest residual measurement of school-textbook inventory leftovers against live Release pointer `lexicon-manifest-2a35f2d32722` (20,112 entries; `poc_thin` 5,499, `old_gate_no_english_anchor` 482, `search_no_visible_gloss` 5).

PR **#6809 is CLOSED**. Issue **#6371 stays OPEN**. No manifest apply/promote, no pointer PR, no `--allow-richness-regression`, no `make practice-deck`.

---

### 1. Residual Summary

| class | n |
|---|---:|
| textbook inventory approved rows | 24,682 |
| already in live Atlas | 16,932 |
| VESUM-ok missing from live | 7,750 |
| VESUM absent | 0 |
| of missing, already have EN in inventory/enrich dry-run | 1,514 (inventory) / 4,592–4,738 (dry-run) |

---

### 2. Missing Candidate Breakdown (7,750 proposed additions)

| Metric | In Inventory | After Local Enrich Dry-Run | Share |
|---|---:|---:|---:|
| **Total proposed additions** | 7,750 | 7,750 | 100.0% |
| **VESUM attested** | 7,750 | 7,750 | 100.0% |
| **VESUM absent** | 0 | 0 | 0.0% |
| **Sourced translations filled** | 0 | 3,806 – 4,022 | 49.1% – 51.9% |
| **Entries with English anchor** | 1,514 | 4,592 – 4,738 | 59.3% – 61.1% |
| **Residual lacking English anchor** | 6,236 | 3,012 – 3,158 | 38.9% – 40.7% |
| **Richness-OK ( $\ge$ 5 rich sections)** | 0 | 0 | 0.0% |
| **`poc_thin` entries** | 7,750 | 7,750 | 100.0% |

---

### 3. Tool Execution & Audit Evidence

[...]

PR #7235, opened 2026-08-24
fix(lexicon): cached-slovnyk-only must not live-fetch; add --no-pointer (#6371)

krisztiankoos:
## Summary

While measuring #6371 fillability, `--cached-slovnyk-only` still called `_slovnyk_cache()` (live HTTP). That blocked a local estimate (~4s/word). This PR:

- Loads the on-disk schema-v4 cache file directly when `cached_slovnyk_only` is set (headword and fallback base).
- Adds `--no-pointer` so `--write` can update a worktree manifest without touching the release pointer.
- Tests for `--no-pointer`.

Does **not** undraft #6809, does not flip the pointer, does not promote the 7750.

Fillability result (local sources): 4022 EN fills, **0** richness-ok, union `poc_thin` 8804→12826 (would fail the publish gate). Tracked on #6371.

## Test

`.venv/bin/python -m pytest tests/test_reenrich_thin_manifest_entries.py -q`

X-Agent: grok-atlas (PR opened from agy/atlas-6371-fillability)

Refs #6371

### DIFF
--- a/scripts/lexicon/reenrich_thin_manifest_entries.py
+++ b/scripts/lexicon/reenrich_thin_manifest_entries.py
@@ -429,7 +429,11 @@ def _translation_for_entry(
     lemma = str(entry.get("lemma") or "")
     entry_pos = entry.get("pos")
     gloss_hints = enrich_manifest._surface_gloss_hints(entry)
-    slovnyk_cache = enrich_manifest._slovnyk_cache(lemma)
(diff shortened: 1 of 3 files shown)


krisztiankoos (commented):
Fleet CF: APPROVE from claude-sonnet-5 at ebe0ee24 (review-7235-r2). Relayed by grok-atlas.

PR #7244, opened 2026-08-24
fix(dispatch): harden branch-holder release — unparseable-state refusal + refusal-record clobber guard (#7242)

### LINKED ISSUE #7242: [infra][dispatch] Harden the #7236 branch-holder release predicate: corrupt-record-as-refusal, refusal-record clobber race, cwd TOCTOU
Follow-up to PR #7241 (#7236). CF review of record (kimi, PASS) identified three Low residuals — none reachable without an already-degraded environment, recorded verbatim below. Priority order:

1. **Corrupt bound task record must read as REFUSAL, not absent** — the one spot where uncertainty currently degrades toward releasable; cheap fix (treat exists-but-unparseable state file as a refusal reason).
2. **Refusal-record write can clobber a concurrent live record** for the same task-id (guard→write race); a pre-write re-check or per-task-id lock closes it.
3. **TOCTOU on the live-cwd probe** — acknowledged accepted residual; document it beside the existing dirty-TOCTOU comment at minimum.

1. **(Low) Corrupt bound task record reads as "absent" → releasable.** `_read_state` returns `None` on `JSONDecodeError` (`scripts/delegate.py:574-578`); `_task_state_for_worktree` silently skips such files (`scripts/delegate.py:2217-2219`) and the candidate re-read in `_branch_holder_activity_reason` also skips non-dict states (`scripts/delegate.py:2340-2341`). A corrupt record for a genuinely RUNNING task is then invisible to the status/pid checks, and release rests on the two external probes (monitor API + lsof) plus clean+synced. Under the P0-reaper posture this is the one place uncertainty degrades toward releasable instead of refusal. Practical blast radius is small (release still requires clean tree + HEAD==origin + dead/absent probes, and a truly live worker's cwd is almost always inside the tree so lsof catches it), but a cheap hardening exists: treat "state file exists but unparseable" as a refusal reason. Non-blocking; follow-up candidate.

2. **(Low) Refusal record can clobber a concurrent live record for the same task-id.** `_record_worktree_prep_failure` writes unconditionally via `_write_state_atomic(_state_path(task_id), ...)` (`scripts/delegate.py:5135`). The #6980 duplicate guard (`scripts/delegate.py:5326-5368`) runs earlier and covers the only live statuses ("running"/"spawning" — verified by grep; "queued"/"starting" do not exist in this codebase), and `--force-new` correctly refuses a live pid and archives (moves, never overwrites — `_archive_task_artifacts`, `scripts/delegate.py:395-412`). The residual hole is a race: two concurrent dispatches with the same task-id both pass the guard (no record yet, no lock held between guard and write); dispatch A spawns and writes status=spawning; dispatch B hits a prep refusal and overwrites A's record with status=failed. Since the auto worktree path derives from agent+task-id, B's failed record's `worktree_path` binds to A's live worktree, making it releasable. Requires concurrent duplicate task-id dispatch (operator error the guard is designed to deter but not make atomic). Non-blocking; a pre-write re-check or per-task-id lock would close it.

3. **(Low) TOCTOU on the live-cwd probe.** The lsof snapshot (`reap_worktrees._live_cwd_paths`, `scripts/orchestration/reap_worktrees.py:421-437`) is taken inside `_branch_holder_activity_reason`; a process that cd's into the holder between the check and `git worktree remove` is not caught — git's no-force refusal covers only dirty/untracked content, not a clean tree in use. The code comment at `scripts/delegate.py:2419-2420` acknowledges the dirty TOCTOU but not the cwd one. Hard to close atomically; noted as accepted residual.

krisztiankoos:
Closes #7242 (the three CF findings from PR #7241's review of record, hardening #7236).

- **Unparseable-state refusal:** a layout-bound task state file that exists but fails to parse now REFUSES release ("uncertainty never reads as releasable" — P0-reaper posture). Previously it read as absent and could degrade toward releasable.
- **Refusal-record clobber guard:** `_record_worktree_prep_failure` re-checks before writing and refuses to overwrite a concurrent `running`/`spawning` record (closes the guard→write race from finding 2).
- **cwd TOCTOU documented** beside the existing dirty-TOCTOU comment (accepted residual, finding 3 — no code).

## Verification
- Worker (cursor, attested Composer 2.5, task `release-hardening-7242`): both new tests mutation-checked (guard bypassed → test fails; restored → passes); full delegate probe 414 passed; ruff clean.
- Driver independent re-probe: same 8-suite probe → **414 passed, 0 failed**; commit scope 2 files.

https://claude.ai/code/session_01LKAACyNj7cxjPsdsFw5Lho

### DIFF
--- a/scripts/delegate.py
+++ b/scripts/delegate.py
@@ -586,6 +586,20 @@ def _read_state(path: Path) -> dict[str, Any] | None:
     return _hydrate_read_only_checkout_snapshots(state)
 
 
+def _state_file_unparseable(path: Path) -> bool:
+    """True when a state file exists on disk but cannot be parsed to a dict."""
+    return path.is_file() and _read_state_json(path) is None
+
+
+def _bound_task_state_unparseable_reason(path: Path) -> str | None:
+    """Return a refusal reason when a layout-bound state file is corrupt."""
+    for task_id in _task_status_candidates_for_worktree(path):
+        state_path = _state_path(task_id)
+        if _state_file_unparseable(state_path):
+            return f"unparseable task state for task-id={task_id}"
+    return None
+
+
 def _pid_alive(pid: int) -> bool:
     """Check if a PID is currently alive via signal-0 probe.
 
@@ -2320,6 +2334,10 @@ def _branch_holder_activity_reason(
     if not candidate_task_ids:
         return "holder is outside the dispatch layout"
 
+    unparseable = _bound_task_state_unparseable_reason(path)
+    if unparseable is not None:
+        return unparseable
+
     try:
         from scripts.orchestration import reap_worktrees
 
@@ -2373,6 +2391,9 @@ def _stale_branch_holder_releasable(path: Path, branch: str) -> tuple[bool, str]
         return False, "dirty"
     if not _worktree_matches_origin_branch(path, branch):
         return False, "HEAD != origin/<branch>"
+    unparseable = _bound_task_state_unparseable_reason(path)
+    if unparseable is not None:
+        return False, unparseable
     task_id, task_state = _task_state_for_worktree(path)
     activity = _branch_holder_activity_reason(
         path,
@@ -2417,7 +2438,9 @@ def _release_stale_branch_holders(
             released.append(path)
             continue
         # No --force: if the tree went dirty after the cleanliness check
-        # (TOCTOU), git refuses removal and we leave the holder mounted (#5708 CF).
+        # (dirty-TOCTOU), or a process adopted the holder as cwd after the
+        # live-CWD probe (live-cwd TOCTOU — accepted residual), git refuses
+        # removal and we leave the holder mounted (#5708 CF, #7242).
         try:
             proc = subprocess.run(
                 ["git", "worktree", "remove", str(path)],
@@ -5063,8 +5086,12 @@ def _record_worktree_prep_failure(
     silence_timeout: float | None = None,
     initial_response_timeout: float | None = None,
     max_budget_usd: float | None = None,
-) -> None:
-    """Persist a terminal failed task record when worktree provisioning is refused."""
+) -> bool:
+    """Persist a terminal failed task record when worktree provisioning is refused.
+
+    Returns True when a record was written. Refuses to overwrite an existing
+    running/spawning record (pre-write re-check closes the guard→write race).
+    """
     if str(_REPO_ROOT / "scripts") not in sys.path:
         sys.path.insert(0, str(_REPO_ROOT / "scripts"))
     from agent_runtime.telemetry import resolve_dispatch_start_telemetry
@@ -5132,7 +5159,12 @@ def _record_worktree_prep_failure(
         failed_state["harness"] = requested_harness
     if lifecycle_carrier is not None:
         failed_state["task_lifecycle"] = lifecycle_carrier
-    _write_state_atomic(_state_path(task_id), failed_state)
+    state_path = _state_path(task_id)
+    existing = _read_state(state_path)
+    if existing is not None and existing.get("status") in ("running", "spawning"):
+        return False
+    _write_state_atomic(state_path, failed_state)
+    return True
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_delegate.py
+++ b/tests/test_delegate.py
@@ -9381,3 +9381,98 @@ def test_cmd_dispatch_refusal_on_base_resolution_writes_terminal_task_record(
     assert state["returncode_reason"] == "worktree preparation failed"
     assert "could not fetch origin/non-existent-base-branch" in (state["last_error"] or "")
     assert state["finished_at"] is not None
+
+
+# --- Issue #7242: Branch-holder release hardening ---
+
+
+def test_branch_holder_refuses_unparseable_bound_task_state(tmp_path, monkeypatch, tmp_tasks_dir):
+    """#7242: exists-but-unparseable bound state must refuse release (P0-reaper posture)."""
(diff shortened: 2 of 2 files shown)


krisztiankoos:
VERDICT: **PASS (approve)** — PR #7244 closes the three #7242 findings I raised on PR #7241. Two narrow, non-blocking residuals noted below (findings 2 and 4); neither warrants re-review.

Exact head reviewed: `32c65d43cbc05ca69a1eb80728d5e7ebcdd6d845` (1 commit on base `main`, `scripts/delegate.py` + `tests/test_delegate.py` only, +131/−4). Tree clean before and after all checks; no edits/commits/pushes left behind.

**1. (Finding 1 — unparseable bound state → refusal): CLOSED for the canonical dispatch layout.**
`_state_file_unparseable` (scripts/delegate.py:589-591) distinguishes exists-but-corrupt from absent via `path.is_file() and _read_state_json(path) is None`; `_bound_task_state_unparseable_reason` (:594-600) enumerates `_task_status_candidates_for_worktree` and refuses. Installed at BOTH release-decision sites: `_stale_branch_holder_releasable` (:2394-2396, before `_task_state_for_worktree`) and `_branch_holder_activity_reason` (:2337-2339, before the candidate re-read loop). All release decisions funnel through these (`_release_stale_branch_holders` :2426, dry-run re-check :3878; no callers outside delegate.py). The candidate re-read loop from my finding (base :2340-2341, now :2357-2359) is fully covered — the guard enumerates the identical candidate set and runs first. `_read_state_json` itself (base :574-578) is deliberately unchanged; the new helper layers the distinction on top. Correct design.

**2. Residual on finding 1 (non-blocking): the worktree_path-bound scan (base :2217-2219, now :2228-2242) still silently `continue`s on unparseable records (:2231-2233).** The guard compensates only for state filenames that round-trip from the flattened path component. `_auto_worktree_path` flattens `[^A-Za-z0-9._-]+` → `-` (:1179), so a task_id containing `/` (e.g. `foo/bar` → path `foo-bar`, state file `foo_bar.json`) produces candidates {`foo-bar`, `agent-foo-bar`, `agent_foo-bar`} — none maps to `foo_bar.json`. A corrupt record of that form bound via `worktree_path` remains invisible to scan AND guard, and with both liveness probes empty the holder releases as "task record absent". Narrow: needs corrupt state + non-round-tripping task_id + empty active-task and lsof probes. Recommend a follow-up note or test; not a re-review trigger.

**3. (Finding 2 — clobber guard): CLOSED per my own prescription.**
Pre-write re-check at scripts/delegate.py:5162-5165 (`_read_state` → refuse if status ∈ {running, spawning}) is exactly the "pre-write re-check" option I prescribed. The residual re-check→write window is in-function and microscopic against an atomic-rename write — acceptably narrow. Doc nit: the docstring (:5089-5093) claims it "closes the guard→write race" but does not label the re-check→write window an accepted residual, unlike the cwd-TOCTOU comment (:2440-2443) which does. Also observed: an existing *unparseable* record reads as None and gets overwritten — benign (atomic writes keep live running records parseable) but asymmetric with the release path's refuse-on-corrupt posture. Callers (:5521, :5765, :5813) ignore the new bool return, which is fine — they print the refusal and exit 1 either way.

**4. (Finding 3 — cwd-TOCTOU documented): CLOSED.** scripts/delegate.py:2440-2443 now explicitly names both the dirty-TOCTOU and the "live-cwd TOCTOU — accepted residual", citing #5708 CF and #7242.

**5. Mutation checks — I ran both source mutants myself; each test kills its claimed mutant:**
- Mutant 1: inserted `return None  # MUTANT` at the top of `_bound_task_state_unparseable_reason` →
  `tests/test_delegate.py:9406: in test_branch_holder_refuses_unparseable_bound_task_state / assert releasable is False / E assert True is False` → `1 failed, 304 deselected`.
- Mutant 2: neutralized the re-check with `if False and existing is not None ...  # MUTANT` in `_record_worktree_prep_failure` →
  `tests/test_delegate.py:9446: in test_record_worktree_prep_failure_refuses_clobber_running_record / assert wrote is False / E assert True is False` → `1 failed, 304 deselected`.
- Restored via `git checkout -- scripts/delegate.py` after each; post-restore both tests pass (`2 passed, 303 deselected`), `git status --short` empty. The tests also embed their own monkeypatch mutation checks with restore-confirmation — good hygiene.

**6. Legitimate release/refusal paths unchanged — full probe green on a verified-clean tree:**
`414 passed in 72.63s` for the driver's exact 8-file command (`tests/test_delegate.py tests/test_delegate_readonly_guard.py tests/test_delegate_runtime_tmp_sweep.py tests/test_delegate_ownership.py tests/test_delegate_primary_integrity.py tests/test_delegate_check_budget.py tests/test_delegate_timeouts.py tests/test_subprocess_timeout_guard.py`), run with `/Users/krisztiankoos/projects/learn-ukrainian/.venv/bin/pytest -q` from the worktree root at head `32c65d43cbc05ca69a1eb80728d5e7ebcdd6d845`. Matches the driver's 414. (An earlier background run also showed 414 passed but overlapped my mutation window, so I re-ran it clean; the quoted result is the uncontaminated one.)

DELIVERABLE: {"outcome":"no_change","reason":"read-only cross-family review of PR #7244 completed; verdict PASS with two non-blocking residual observations; mutation checks restored and tree verified clean"}

_Review of record: kimi (Moonshot K3) — the same seat that raised the #7242 findings — cross-family vs author cursor (attested Composer 2.5); task review-7244, head 32c65d43cb, both source mutants executed. Posted by the infra driver. The finding-2 residual (unparseable record bound via a non-round-tripping task_id, needs a triple-degraded environment) is recorded on #7242 as a footnote, not re-opened._

Issue #7177, opened 2026-08-23
[EPIC][monitor] Monitor API + UI — one glance at epics, host, AI workers, and project state


krisztiankoos:
Area: **infra** · stream `monitor`. Operator order 2026-08-23; **one-server hygiene 2026-09-02** (grind-3 wording scrub).

## Why a dedicated epic
The local Monitor API (`localhost:8765` via loopback/tunnel; code in `scripts/api/`) and dashboards (`dashboards/*.html`) are how developers and orchestrators see the fleet. This epic owns making Monitor the single source of truth: remote epic lifecycle, host/worker/project glance, observer honesty, OPSEC.

## User outcome
A developer or orchestrator — with no SSH — opens ONE Monitor page and within 5 seconds knows:
1. **Epics** — every epic, who drives it (agent, harness, opaque host), lease health, latest handoff; claim/resume from anywhere that reaches the API.
2. **Hosts** — the production host and optional Mac observer (opaque ids only); load/RAM/disk; `unknown` never shown as idle.
3. **AI workers** — live workers, driver seats, paid lanes (lane, epic, task id, age, state), each exactly once.
4. **Projects** — checkouts, serving SHA vs `origin/main` (drift), dirty tree, worktrees, services up/down.
5. **Attention** — broken/stale/flapping with honest empty states.

## Topology (locked)

- **One production Linux host** runs Monitor/Sources/Work/Astro/teacher as deployed. Teacher/Hramatka stays hosted — do not destage.
- **Mac** is operator laptop / observer (tunnel client), **not** a second fleet host.
- **Gone / do not plan against:** a second production Linux writer, deleted runner hosts, Patroni/HA, k8s.

## Milestones (status-aware)

**M1 — Remote epic lifecycle v1.** API-host store is authority for claim/heartbeat/handoff/release; TTL+heartbeat only; opaque host id. Design: #7178.
**M2 — Test + migrate.** Held-out blind resume Mac↔production host via API (#7184); seed registry (#7185); launchers on production host + Mac fail-closed to API; local store demoted.
**M3 — Epic overview UI.** `dashboards/epics.html`.
**M4 — Host / worker / project glance.** Occupancy honesty (#7491); project-state reporter on production host + Mac (#7257); finish related glance children.
**M5 — Observer reliability + OPSEC.** Tunnel health; honest freshness; sanitizer over public routes (no IPs/aliases/paths).

## Denominator

[...]

PR #7246, opened 2026-08-24
feat(epics): launcher auto-resolves its own opaque host_id (#7229)

### LINKED ISSUE #7229: [monitor] launcher auto-resolves its own opaque host_id (stop defaulting driver leases to 'local')
Parent: #7177 (M2 residual). Cosmetic-but-worth-it: driver leases currently show holder `host_id=local` in `/api/epics/v1` and epics.html because `scripts/lib/session_supervisor.sh` defaults `LU_MONITOR_HOST_ID` to `local`. Make each launcher self-identify instead of requiring a manual per-host env.

## Outcome
When `LU_MONITOR_HOST_ID` is unset, `claim_session_supervisor_env` resolves the host's OWN opaque id from the canonical mapping already used by occupancy (`MONITOR_OCCUPANCY_HOST_IDS`, e.g. `hramatka=host-teacher,atlas-runner=host-job`; Mac=`mac-operator`) via `scripts/api/occupancy_local` self-host resolution (`self_host_opaque_ids` / `ENV_DRIVER_HOST_ID`). Fall back to `local` ONLY when it genuinely cannot resolve (fail-safe, never guess/leak). An explicit `LU_MONITOR_HOST_ID` still wins.

## Constraints
- Opaque ids only; never derive/print a hostname, IP, SSH alias, or path.
- No behavior change when the env is already set or when resolution fails (stays `local`).

## Acceptance
- On a host with the canonical mapping, a driver claim shows the right opaque host in `/api/epics/v1` (unit test with a fixture mapping + fake hostname; no live SSH).
- Explicit env override still honored; unresolvable → `local`.
- `bash -n` + shellcheck clean; `ruff` clean on any Python touched.

krisztiankoos:
Parent: #7177. Closes #7229. Authored by codex lane (task monitor-host-id-autoresolve); PR opened by main (worker gh unauth).

Launcher self-identifies its opaque host_id instead of defaulting driver leases to `local`:
- canonical opaque-host resolver (reuses occupancy self-host mapping: `hramatka=host-teacher, atlas-runner=host-job`, Mac=`mac-operator`)
- explicit `LU_MONITOR_HOST_ID` still wins; unresolvable → `local` (fail-safe)
- tests: resolver + explicit-override + failure-fallback, mutation-sensitive

Worker evidence: 45 tests passed; Ruff + `bash -n` + ShellCheck clean.
**NOT yet independently reviewed by main** — needs cross-family review + own verification before merge (next session). No auto-merge.

### DIFF
--- a/scripts/api/occupancy_local.py
+++ b/scripts/api/occupancy_local.py
@@ -399,9 +399,31 @@ def foundry_marker_host_id(mapping: dict[str, str]) -> str | None:
     return None
 
 
+def resolve_launcher_host_id() -> str:
+    """Resolve the launcher host id from the canonical occupancy configuration."""
+    explicit = os.environ.get("LU_MONITOR_HOST_ID")
+    if explicit is not None:
+        return explicit or "local"
+
+    driver_host_id = os.environ.get(ENV_DRIVER_HOST_ID, "").strip().lower()
+    if driver_host_id and _opaque_host_id(driver_host_id):
+        return driver_host_id
+
+    try:
+        from scripts.api.occupancy import parse_host_id_map  # noqa: PLC0415 — breaks the occupancy import cycle
+
+        claimed = self_host_opaque_ids(parse_host_id_map())
+    except Exception:
+        return "local"
+    if len(claimed) == 1:
+        return next(iter(claimed))
+    return "local"
+
+
 def _build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Publish or clear a sanitized occupancy marker (opaque host_id only).")
     sub = parser.add_subparsers(dest="command", required=True)
+    sub.add_parser("resolve-host-id", help="resolve the current process's opaque occupancy host id")
     heartbeat = sub.add_parser("heartbeat", help="write one marker")
     heartbeat.add_argument("--kind", required=True)
     heartbeat.add_argument("--task-id", required=True)
@@ -417,6 +439,9 @@ def _build_parser() -> argparse.ArgumentParser:
 
 def main(argv: list[str] | None = None) -> int:
     args = _build_parser().parse_args(argv)
+    if args.command == "resolve-host-id":
+        print(resolve_launcher_host_id())
+        return 0
     if args.command == "heartbeat":
         written = write_marker(
             kind=args.kind,
--- a/scripts/api/occupancy_local.py
+++ b/scripts/api/occupancy_local.py
@@ -410,7 +410,7 @@ def resolve_launcher_host_id() -> str:
         return driver_host_id
 
     try:
-        from scripts.api.occupancy import parse_host_id_map  # noqa: PLC0415 — breaks the occupancy import cycle
+        from scripts.api.occupancy import parse_host_id_map  # noqa: PLC0415 — # lazy-ok: occupancy cycle breaker
 
         claimed = self_host_opaque_ids(parse_host_id_map())
     except Exception:
--- a/scripts/lib/session_supervisor.sh
+++ b/scripts/lib/session_supervisor.sh
@@ -126,7 +126,15 @@ claim_session_supervisor_env() {
   # Remote v1 uses the design-note TTL; liveness is carried by the independent
   # launcher renew loop, never by a server-side PID probe.
   local ttl_seconds=900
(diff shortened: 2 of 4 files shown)


krisztiankoos:
Main (Claude seat) independent verification at head `d8231ce59d` — detached worktree, primary venv, primary stayed on main:

- `ruff check` on the 3 Python files: clean · `bash -n scripts/lib/session_supervisor.sh`: ok
- `pytest tests/api/test_occupancy_local.py tests/test_session_supervisor_shell.py`: **13 passed**
- Mutation checks (#M-16): forcing the shell resolver branch to `host_id="local"` → `test_claim_resolves_unset_host_id_and_preserves_explicit_override` **fails**; disabling the single-claimed-host return in `resolve_launcher_host_id` → `test_resolve_launcher_host_id_uses_occupancy_mapping_and_fallbacks` **fails**. Both guards are real.
- Diff read: precedence explicit env (set-but-empty → `local`, shell and Python agree) → `MONITOR_OCCUPANCY_DRIVER_HOST_ID` → occupancy self-host mapping → `local`; resolver stderr is swallowed and a non-zero exit falls back to `local`, so `claim_session_supervisor_env` cannot break; `-m scripts.api.occupancy_local` uses the same `$python_bin`/cwd assumption the function already makes for `scripts.session_supervisor`.

Cross-family review of record (Gemini seat, task `review-7246-agy`) is running; verdict will be posted here before merge.

krisztiankoos:
VERDICT: APPROVE — cross-family review of record (Gemini seat, `gemini-3.7-flash` via agy, task `review-7246-agy`; author = codex).

Reviewer evidence (from the task result): `git diff origin/main...HEAD` read; 13 targeted + 45 occupancy/supervisor tests pass; ruff clean; shellcheck clean; both import orders of `scripts.api.occupancy` ↔ `occupancy_local` load (lazy import avoids the cycle); fail-safe path simulated under `set -euo pipefail` → `host_id=local`. Findings: nits only, no changes requested.

Gate: CF review APPROVE + main verification (comment above) + required CI green → arming auto-merge (squash).

krisztiankoos:
Merge-queue kick root-caused (infra lane, awareness for the owning codex/monitor lane — run 32737542576, shard 3/4): `tests/api/test_import_pinning.py::test_ast_imports_and_loaders` fails with "Uncovered nested import 'scripts.api.occupancy' in scripts/api/occupancy_local.py:413" — that nested import is introduced by this PR (main at 4dd731792e has no such import and the test passes there: 5 passed). The PR tier never ran this test (it is not a changed-file test), so the PR stayed green until the merge queue's full shard caught it. Fix in the authoring lane: cover the import per the test's instruction (PRELOAD_MODULES/OPTIONAL_MODULES or the documented lazy-import exemption), run `tests/api/test_import_pinning.py` + `tests/api/test_occupancy_local.py`, exact-head CF re-stamp, re-queue. If this is still sitting in an hour the infra lane will dispatch the delta.

krisztiankoos:
Removed from the merge queue at 14:26Z: the merge-group run (full pytest, run 32737542576) fails
`tests/api/test_import_pinning.py::test_ast_imports_and_loaders` — the lazy `from scripts.api.occupancy import parse_host_id_map` in `scripts/api/occupancy_local.py:413` is an uncovered nested import for the API process (guard passes on `main`; reproduced locally at `d8231ce59d`). Both reviews ran targeted tests only, which is why it slipped — the PR-tier fastlane does not run this guard.

Fix dispatched back to the authoring lane (codex, task `monitor-host-id-pinning-fix`); I will re-verify the delta and re-queue. Not merging until the full guard is green.

krisztiankoos:
Delta `deb2570e03` (codex lane, task `monitor-host-id-pinning-fix`): one-line change — the deferred import is marked `# lazy-ok: occupancy cycle breaker`, the guard's documented exemption; preloading `scripts.api.occupancy` would be wrong because it imports `occupancy_local` at module top level (the cycle is why the import is nested).

Main verification at `deb2570e03` (detached worktree, primary venv): `pytest tests/api/test_import_pinning.py tests/api/test_occupancy_local.py tests/test_session_supervisor_shell.py` → **18 passed**; ruff clean; both import orders load. APPROVE stands; auto-merge re-armed so the PR re-enters the queue once PR-tier checks pass.

Issue #7269, opened 2026-08-24
[monitor][design][decision] Monitor API app factory + typed root/store context — hardening proposed by the OPSEC-sweep critique (needs operator/advisor go)


krisztiankoos:
Parent: #7177 (M5). Refs #7182 (design r2/r3 critique, task `critique-7182-design-r2-codex`).

**What was proposed (by the GPT-seat critic, not yet decided).** The Monitor API is a module-global `app` with an attached `_lifespan` (`scripts/api/main.py`), and roots/stores are import-bound module globals (`scripts/api/config.py`: `PROJECT_ROOT`, `LIVE_REPO_ROOT`, `MESSAGE_DB`, dashboard/batch roots; per-router `_repo_root/_db_path/_store` globals; global in-memory presence/report stores). The critic recommends a `create_app(..., lifespan=None)` factory plus a typed app-scoped root/store context propagated to routers, so tests and tools can build fully isolated instances without monkeypatching, and so no route can reach a real store by accident.

**Why it is a decision, not a task.** It changes how the API is composed (every router's access to roots/stores), touches most of `scripts/api/`, and sets the pattern for all future routers. Under the operator contract this is new architecture and needs an explicit go from the operator or an advisor before any lane implements it.

**What proceeds without it.** #7182 PR-A (the sweep harness) uses the existing patchable seams + `TestClient(app)` without the lifespan, so the OPSEC sweep does not block on this.

**Recommendation (main orchestrator).** Approve as a follow-on hardening after the #7182 burn-down: factory + context introduced incrementally per router family behind the existing seams, with the sweep as the regression net. Reject only if the operator wants the API composition left as-is; then the seams stay test-only.

krisztiankoos:
**APPROVED (operator 2026-08-25).** Design sketch below for the implementing driver to formalize into
`docs/design/monitor-api-app-factory.md` + per-family issues. Regression net: the OPSEC route sweep
(`tests/api/opsec_sweep/`) — every step keeps it green and DELETES the seam it makes redundant.

## Design of record (sketch)
1. **`MonitorContext` (frozen dataclass)** — one object holding every root and store handle: `roots`
   (project_root, live_repo_root, dashboards_dir, batch_state_dir, curriculum_root, plans_root,
   backup_dir, logs_dir, queue_dir, pid_dir, docs EFFECTIVE_ROOTS, image roots…) and `stores`
[...]

PR #7296, opened 2026-08-25
docs(monitor): formalize Monitor API app-factory design (#7269)

krisztiankoos:
## What
Formalizes the operator-approved design sketch (posted on #7269, 2026-08-25) into
`docs/design/monitor-api-app-factory.md`: `MonitorContext` + `create_app()` factory, the
incremental one-router-family-per-PR rollout (step 0 core no-op through step 6 "the rest"),
constraints, the OPSEC sweep as regression net, and the sequencing this driving session follows.

No code changes — docs only. Content is a formalization of the approved sketch, not new design
judgment (see #7269 comment for the source of record).

## Why
First driver task per the #7177 epic handoff (2026-08-25 08:36:42Z comment): formalize the sketch
→ get one cross-family design review → file per-family sub-issues under #7269 → implement step 0→6.
This PR is step 1 of that sequence.

## Review
Requesting one cross-family design review (codex seat) per the epic handoff routing note. Will
file per-family sub-issues under #7269 once review lands, then dispatch step 0 (core no-op
refactor) implementation.

Refs #7177, #7269.

### DIFF
--- a/docs/design/monitor-api-app-factory.md
+++ b/docs/design/monitor-api-app-factory.md
@@ -1,24 +1,25 @@
 # Monitor API App Factory + Typed Root/Store Context — Design Doc
 
-> **Version:** v6 (revised after five rounds of cross-family review)
+> **Version:** v7 (revised after six rounds of cross-family review)
 > **Status:** APPROVED (operator, 2026-08-25) — hardening, not new architecture; proceeds without
-> further sign-off. Rollout is incremental and gated per family below. **Not yet re-reviewed** — v6
-> needs a sixth cross-family pass before step 0 is dispatched (see § 10).
+> further sign-off. Rollout is incremental and gated per family below. **Not yet re-reviewed** — v7
+> needs a seventh cross-family pass before step 0 is dispatched (see § 10).
 > **Author:** Claude (monitor-epic driver), formalizing the operator's sketch, then revising per
-> five rounds of codex review below.
+> six rounds of codex review below.
 > **Issue:** #7269 (parent epic #7177, milestone M5 hardening).
 > **Refs:** #7182 (OPSEC sweep design r2/r3 critique, task `critique-7182-design-r2-codex`, GPT-seat
 > critic origin of the proposal).
-> **Changelog:** v1→v5 each got a cross-family REJECT and fixed the prior round's findings — see §10
-> for the full five-round record. Round 5 confirmed the v5 middleware mechanism itself sound (it does
-> avoid the round-4 identity-equality problem) and confirmed the CORS/resilience/lifespan/handler
-> wiring otherwise matched `main.py`, but found the exception-handler snapshot incomplete — a bare
-> `FastAPI()` app registers 3 default handlers, one of which (`WebSocketRequestValidationError`)
-> `main.py` never overrides, so the real app has 4 exception-handler keys, not the 3 v5 claimed — and
-> flagged that the `repr()` fallback v5 proposed for the middleware comparison is unstable (Starlette's
-> middleware `repr()` embeds a function memory address). v6 (this version) fixes both, with every
-> value in §5.1 point 3 now independently confirmed by importing the real `scripts.api.main.app` and
-> inspecting it directly, not assumed — see §10's v5 entry.
+> **Changelog:** v1→v6 each got a cross-family REJECT and fixed the prior round's findings — see §10
+> for the full six-round record. Round 6 confirmed v6's 4-key exception-handler set and its
+> `.cls`/`.args`/`.kwargs` middleware attributes were both correct, but found two more issues: v6's
+> closing sentence for the exception-handler bullet stated the per-key custom/default origins
+> backwards from what the bullet's own preceding list already said correctly (a self-contradiction),
+> and the lifespan check was underspecified — importing the real app shows
+> `app.router.lifespan_context` is FastAPI's `merged_lifespan` wrapper, not `_lifespan` itself, so an
+> identity check against `_lifespan` would never pass. v7 (this version) fixes both — the
+> exception-handler bullet now states each key's origin explicitly rather than positionally, and the
+> lifespan check is redefined to call `_lifespan` directly (reusing an existing test pattern already
+> in this codebase) instead of comparing the wrapped object — see §10's v6 entry.
 
 ---
 
@@ -176,7 +177,7 @@ by itself is not precise enough to file a bounded sub-issue from, and lumping se
 worse. Guessing a finer split without reading what's actually inside those six files would just move
 the same ambiguity to a smaller-looking table.**
 
-### 5.1 Step 0 — Core (revised after the v2 through v5 reviews — see §10)
+### 5.1 Step 0 — Core (revised after the v2 through v6 reviews — see §10)
 
 `MonitorContext` + `create_app` + `production_context`/`fixture_context` per §4. **No router
 changes** — every router still reads its own module globals directly, exactly as today. This matters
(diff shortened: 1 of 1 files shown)


krisztiankoos:
**Cross-family review (codex, verdict: REJECT)** on the v1 commit — 3 P1 + 1 P2 findings:
1. Rollout families (step 1–6) named concepts, not exact routers; ambiguous enough that a sub-issue
   couldn't be filed from them directly.
2. OPSEC-sweep-as-regression-net had no seam inventory/baseline/strict-decrease assertion.
3. `fixture_context`'s outside-root guard was described as structural but nothing in the design
   actually prevented an arbitrary `sqlite3.connect` bypassing it.
4. Step 0's "byte-identical outcome" had no executable definition.

**v2 pushed** (5d3a6c4) addressing all four — grounded in `main.py`'s live 33-router
`include_router` list (pulled directly, not guessed): per-router-file steps for the six 1,000+ line
multi-concern routers, a new mechanical "router dependency & seam inventory" step that the rest of
the family boundaries derive from, a concrete `_open_db` choke-point design for the outside-root
guard, and an executable byte-identical-outcome test definition. Full disposition in the doc's new
§10.

Requesting a follow-up cross-family review on v2 before sub-issues get filed under #7269.

krisztiankoos:
**Cross-family review round 2 (codex, verdict: REJECT)** on v2 (commit 5d3a6c4) — 3 P1 + 1 P2:
1. Step 0 claimed a fixture-isolation guarantee no router could actually provide yet (routers still
   read module globals directly at step 0).
2. The byte-identical test was unexecutable as specified — no legacy-app builder, wrong description
   of the OPSEC sweep's denominator (it's 280 HTTP ops across 5 methods + 1 WS route, not "GET/HEAD"),
   no normalization plan for non-deterministic response fields.
3. `_open_db` claimed to be the only DB call site; real access goes through 3 patterns
   (`sqlite3.connect` direct, the existing `connect_sqlite()` wrapper, `SessionStreamDatabase.connect()`)
   across 21 files, and the root-check used unresolved paths (symlink-escapable).
4. The router count itself was wrong — v2 said 33; the real, live-grepped count is 44, and
   `build_events_router` had been silently dropped from the step table.

**v3 pushed** (074b9b4), re-grounded against the live tree (grep-verified, not re-guessed):
step 0 no longer overclaims isolation and reuses two things that already exist (the unchanged
`tests/api` suite + the sweep's existing frozen-denominator digest check) instead of a new
comparison harness; `_open_db` now wraps the existing `connect_sqlite()`/`SessionStreamDatabase`
mechanisms with a 3-pattern grep-lint seeded from a self-checked 21-file count; the router table is
corrected to 44 modules with a per-step tally that sums back to 44. Full disposition in §10.

Requesting a third review round on v3 before step 0 is dispatched for implementation.

krisztiankoos:
**Cross-family review round 3 (codex, verdict: REJECT)** on v3 (commit 074b9b4):
1. §7 (Regression net) still described a fixture-isolation test that §5.1 had already dropped in
   the v2→v3 pass — a real doc-consistency bug on my part.
2. Route-digest + passing tests don't prove app-level wiring (middleware/exception
   handlers/lifespan) is unchanged — a real, previously-unaddressed gap in step 0's no-op proof.
3. Router-count precision: 44 unique modules vs. 47 total mounts wasn't distinguished. (The
   reviewer's specific sub-claim that my grep "misses the multiline reviewer_ghosts_router call"
   did **not** hold up under my own independent re-verification — noted in §10 rather than silently
   accepted.)
4. `connect_sqlite()` caller count was internally inconsistent (said "11 files," listed 10 names).
5. "6 routers are 1,000+ lines" was imprecise for `dashboard_router.py` at 996.

**v4 pushed** (b0361e1) — fixes the doc bug and the real design gap (adds a direct
middleware/exception-handler/lifespan structural-equality check to step 0, alongside the existing
route-digest check), corrects the count precision issues, and records the not-upheld reviewer
sub-claim rather than silently accepting it. Full disposition in §10.

Requesting a fourth review round on v4 before step 0 is dispatched for implementation.

krisztiankoos:
**Cross-family review round 4 (codex, verdict: REJECT)** on v4 (commit b0361e1). Good news first:
this round independently re-verified every round-3 fix rather than trusting the doc, and confirmed
four of them held — §7's isolation claim, the router count (44 unique / 47 mounts), the
`connect_sqlite()` count (10 callers + 1 definition), and the six line counts including
`dashboard_router.py` at 996.

One finding remained: the proposed middleware/exception-handler/lifespan structural-equality check
was itself flawed — it diffed `legacy_app.user_middleware == factory_app.user_middleware`, but
Starlette's `Middleware` entries compare by identity, so two independently-built apps never compare
equal that way even if identically configured; and no `legacy_app` object exists to diff against once
`main.py`'s bottom is rewritten to call the factory.

**v5 pushed** (62ca873) — replaces the two-live-apps diff with a hardcoded expected-snapshot
comparison grounded directly in `main.py`'s actual current construction (read from the file, not
assumed): the factory-built app is checked against literal values (CORS kwargs, the
`resilience_middleware` function, the three exception-handler classes, the `_lifespan` function),
sidestepping identity-equality entirely. Full disposition in §10.

Requesting a fifth review round on v5.

krisztiankoos:
**Cross-family review round 5 (codex, verdict: REJECT)** on v5 (commit 62ca873). Confirmed the fixed
middleware-comparison mechanism itself sound and the CORS/resilience/lifespan/handler wiring
otherwise matched `main.py`. Two findings:
1. **Blocking:** the exception-handler snapshot was incomplete — claimed 3 keys, but a bare
   `FastAPI()` already registers `WebSocketRequestValidationError` as a default handler that
   `main.py` never overrides, so the real app has 4.
2. The proposed `repr()` fallback for middleware comparison was unstable — Starlette's middleware
   `repr()` embeds a function memory address.

**v6 pushed** (5e9f595) — this time verified by importing the real `scripts.api.main.app` directly
and inspecting it live (not just reading `main.py` and inferring): confirmed 4 exception-handler keys
and the exact `entry.cls`/`entry.args`/`entry.kwargs` attribute names, matching what the round-5
reviewer's own isolated probes found. Full disposition in §10.

Requesting a sixth review round on v6.

krisztiankoos:
**Cross-family review round 6 (codex, verdict: REJECT)** on v6 (commit 5e9f595). Confirmed the 4-key
exception-handler set and the middleware `.cls`/`.args`/`.kwargs` attribute names were both correct.
Two findings:
1. v6's closing sentence for the exception-handler bullet stated origins positionally and got them
   backwards from what the bullet's own preceding list already said correctly (self-contradiction).
2. The lifespan check was underspecified — `app.router.lifespan_context` is FastAPI's
   `merged_lifespan` wrapper, not `_lifespan` itself, so an identity check would never pass.

**v7 pushed** (f361012) — exception-handler origins now stated explicitly next to each key (no
positional shorthand), and the lifespan check now reuses an existing pattern already in this codebase
(`tests/api/test_epics_registry_seed.py`, which calls `_lifespan(app)` directly) instead of comparing
the wrapped closure. Full disposition in §10.

Requesting a seventh review round on v7.

krisztiankoos:
**Cross-family review round 7 (codex, verdict: REJECT)** on v7 (commit f361012). Confirmed the
exception-handler origin fix held. Two findings, one the largest across all seven rounds:
1. The lifespan test called `_lifespan(factory_app)` directly, but `_lifespan` ignores its app
   argument entirely — the test could pass even if `create_app()` wired no lifespan at all.
2. **Largest finding:** `main.py` has 18 routes + 3 exception handlers decorated directly on the
   module-level `app` object (not inside any of the 44 router files). The design's "`main.py`'s
   bottom becomes `create_app(...)`" claim can't work as stated — those decorators need `app` to
   already exist where they run, which is impossible if `app` is only built at the bottom.

**v8 pushed** (92d401d) — the lifespan test now uses `TestClient` as a context manager to trigger the
real ASGI lifespan protocol; the 18 inline routes get wrapped into a `core_router` (same pattern as
every other router, included last to preserve route-matching order) with the 3 handlers converted to
explicit `app.add_exception_handler(...)` calls inside `create_app()`. Full disposition in §10.

Requesting an eighth review round on v8.

krisztiankoos:
**Cross-family review round 8 (codex, verdict: REJECT)** on v8 (commit 92d401d). Confirmed the
core-router-last ordering and explicit exception-handler registration both actually work. Three
findings:
1. The lifespan test's positive assertion was optional, not mandatory — a plain `FastAPI()` with no
   lifespan at all would also pass the "no exception" check alone.
2. The inline-route count was wrong *again* (18 claimed across two versions; the real count,
   re-verified via `grep -c`, is 15).
3. Two of those 15 route bodies (`health_check`, `get_config`) read the module-global `app.version`
   directly — would silently break once multiple `create_app()`-built instances need to coexist.

**v9 pushed** (aff2e51) — the lifespan test now requires the spy assertion rather than treating it as
optional, the route count is corrected to 15 throughout, and both handlers switch to
`request.app.version`. Full disposition in §10.

Requesting a ninth review round on v9.

krisztiankoos:
**Cross-family review round 9 (codex, verdict: APPROVE)** on v9 (commit aff2e51). All three round-8
fixes independently re-verified against the checked-out tree: the lifespan spy assertion is
explicitly mandatory, the route count checks out at 15, and the `request.app.version` fix matches
`main.py`'s actual code. "No new blocking findings."

**Design review closed after 9 rounds.** Small final commit (ba92ea3) updates the status line and
review record. Real gaps caught along the way: a router-count arithmetic error (appeared 3 times), a
positional-shorthand self-contradiction, two rounds on a non-functional lifespan test, and the
largest one — main.py has 15 routes decorated directly on the module-level `app` object, which the
original sketch never considered and which breaks the "bottom of the file becomes `create_app(...)`"
claim as originally stated.

Next: filing per-family sub-issues under #7269 (from the §5.2 inventory step's real output, not
before it) and dispatching step 0 implementation.

Issue #6375, opened 2026-08-05
[open-model-data][phase-3-recovery][superseded] Correction/Protection KB and Cycle007 history


krisztiankoos:
Parent stream epic: #6321 (`open-model-data`).

## Recovery classification

The earlier #6333 delivery is retained as historical `ENGINE_READY` evidence. It did not satisfy source coverage or independent linguistic validation and does not complete Phase 3.

**Superseded v1 contract SHA-256:** `6a563a7526c4ec7a89732f3de5651b0ab2e176ec089abf80f9eb733337db7662`

The complete reviewed execution and acceptance contract follows verbatim. Any material change to its outcome, denominators, role separation, evaluation design, breadth floors, completion gates, or boundaries requires a new Ukrainian-domain review and separate scope/circularity review.

---
# GO PHASE 3 RECOVERY — source-derived Ukrainian correction knowledge base

You are the **single accountable root orchestrator** for recovering and completing
Phase 3 of the Ukrainian open-model-data program in
`/Users/krisztiankoos/projects/learn-ukrainian`.

Your job is not to create another demo or five-example rule list. Your job is to
ship a genuinely useful, source-derived, model-neutral correction/protection
knowledge base and consumer tool that Ukrainian NLP and open-weight model teams can
use to prepare training data for clean modern Ukrainian without erasing legitimate
variation, historical language, dialect, folklore, or quoted Russian.

The project is **not training a model in this phase**. It is shipping the means for
teams with compute to train or adapt models better.

## Binding operating contract

1. Read `AGENTS.md`, the full operator contract served by `/api/rules`, and invoke
   the `entire-context` skill before prioritization or dispatch. Use `drive-epic`
   when applicable.
2. Start read-only from the repository root with `git status --short --branch`.
   Keep the primary checkout read-only. Every edit, commit, and PR must happen in
   `.worktrees/dispatch/<agent>/<task>/` on a `codex/`, `grok/`, `agy/`, `claude/`,
   or other valid agent branch with the required `X-Agent` trailer.
3. Run `list_agents` before spawning and after all requested agents are terminal.
   Keep at most three non-root native agents active across the tree. Do not create
   duplicate lanes or shared-file contention.
4. Check CodexBar and live model assignment before fan-out. Route by task fit:
   - Ukrainian extraction and linguistic review: Gemini 3.6 Flash High as the
     primary domain lane; GPT, Grok 4.5, or Claude as independent dissent lanes.
[...]

PR #7304, opened 2026-08-25
fix(open-model-data): support privileged Cycle007 bind mounts

krisztiankoos:
## Outcome

Allow the Cycle007 guardian mount helper to accept an explicit argv vector, enabling a reviewed non-interactive privilege prefix for bind mounts without invoking a shell.

## Safety

- `subprocess.run(..., shell=False)` remains mandatory.
- Empty command vectors and NUL-bearing elements fail closed.
- Provider-free bootstrap and the complete `resume` identity gate are unchanged.
- Zero provider calls.

## Verification

- Ruff passed.
- Guardian and held-out tests: 42 passed.
- Focused guardian/controller/adjudication/Grok suite: 105 passed.
- `git diff --check` passed.

Issue: #6375


krisztiankoos:
Independent cross-family exact-head review at `d17d4590c41bfca013b3a6f51cecf0203202071f`: **APPROVE**. Both prior blockers are fixed: executable paths must be absolute, and the CLI constructs exactly `sudo -n mount` from separate single-token flags. Reviewer confirmed `shell=False`, injection resistance, unchanged provider gates, and 48 passing focused tests.

Issue #7311, opened 2026-08-25
delegate.py needs_finalize check misses committed-but-unpushed dispatches (cursor, 2/2 this session)


krisztiankoos:
Two `--agent cursor --mode danger` dispatches this session (monitor-epic, 2026-08-25) both
finished `status=done` with a clean local commit in the worktree but NOTHING pushed to origin
and no PR opened — the "silent-exit class" flagged in operator memory, now reproduced twice with
a precise root cause.

**Repro:** `monitor-7269-step05-router-inventory` and its follow-up
`monitor-7269-step05-router-inventory-fix1`, both `--agent cursor --mode danger`. Both left a
real commit on the local worktree branch (`333cff007c`, `d07598c8d4`) that never reached
`origin/cursor/monitor-7269-step05-router-inventory` until I pushed it manually.

**Root cause** (`scripts/delegate.py` ~line 4680-4703, the `needs_finalize` gate):

```python
commits_ahead = _count_commits_ahead(Path(worktree_path), base_ref)
if dirty_on_exit in (True, None) and commits_ahead in (0, None):
    needs_finalize = True
```

`_count_commits_ahead` measures commits ahead of the **base branch** (`main`), not ahead of the
dispatch branch's own remote tracking ref. A worker that commits locally but never pushes has
`commits_ahead >= 1` relative to `main` (it looks like real work happened) and a clean working
tree (`dirty_on_exit=False`) — so this condition is false, `needs_finalize` never fires, and the
task settles as plain `done`. The check conflates two different questions: "does this commit
differ from main" (yes) vs. "did this commit ever leave the local worktree" (no).

**Proposed fix:** add a second, independent condition — commits ahead of the branch's own
`@{upstream}` / `origin/<branch>` ref (or: local HEAD SHA not reachable via `git branch -r
--contains <sha>` against the pushed remote) — that also sets `needs_finalize = True` for
write-capable modes. This is orthogonal to the existing base-branch check, not a replacement for
it — both a dirty tree AND a clean-but-unpushed tree are "not actually shipped" states.

**Impact:** every `--mode danger` dispatch's `done` status is currently NOT proof the work
reached origin — the existing driver protocol note ("before declaring a dispatch dead: `gh pr
list` first, then check the worktree for finished-but-unpushed work") is a manual workaround for
exactly this gap, and both instances this session required it.

Not fixing inline — `scripts/delegate.py` is infra-lane-owned tooling; filing for that lane.

[...]

PR #7308, opened 2026-08-25
docs(monitor): draft epics & areas graph map design (#7295)

krisztiankoos:
Parent: #7295 (epic #7177, milestone M5). Not yet implemented — design only, awaiting
cross-family review before #7295's implementation dispatch.

**What this is.** v1 design for `GET /api/epics/graph/v1` (new route on the existing
`epics_router.py`) + `dashboards/epics-map.html`, per #7295's request for a visual
relationship view of areas/epics/issues.

**Real findings from verifying the data sources live (not assumed):**
- A second, coarser area registry exists (`fleet_taxonomy.yaml`) and is **stale** — no
  `monitor` area at all (can't resolve epic #7177, the epic this feature belongs to),
  `infra` still points at the closed #4707 instead of successor #6943. Not reused for
  this design; staleness filed separately as #7306.
- Scoped v1 down to epic→issue **membership** (not a full issue-to-issue mention graph) —
  flagged as an open question for the reviewer, since the issue's "relations" wording
  could be read either way.
- Reuses `issue_stream_audit.py`'s existing rate-limited refresh worker for GitHub calls
  instead of building a second poller.

**Binding constraint carried forward from #7269's review history:** the new route must
update the OPSEC sweep's `FROZEN_HTTP_OPERATION_COUNT` and `FROZEN_DENOMINATOR_SHA256`
together — documented explicitly in §5 so the implementer doesn't get blindsided.

Requesting cross-family review (this is materially smaller scope than #7269's
app-factory doc — budget for iteration, not nine rounds).

X-Agent: claude-monitor

Issue #7306, opened 2026-08-25
fleet_taxonomy.yaml is stale vs issue_streams.yaml (missing monitor #7177, infra points at closed #4707)


krisztiankoos:
Found while designing #7295 (epics/areas graph map, verifying data sources before building on them — not from memory).

`scripts/config/fleet_taxonomy.yaml` (last touched 2026-07-27, `7656697b26`) is a separate, coarser
area registry consumed by `scripts/orchestration/fleet_taxonomy.py`'s `resolve_area()`, which
`scripts/api/work_router.py`'s `/api/work/v1/projection` uses for per-item area tagging.
`scripts/config/issue_streams.yaml` is the finer-grained stream registry (13 streams, 18 epic
slots) and has moved on without a corresponding `fleet_taxonomy.yaml` sync:

- **No `monitor` area at all** — epic #7177 cannot be resolved through `resolve_area()`. Any work
  item tagged against the monitor epic gets no area attribution from `work_router.py`'s
  projection.
- `infra` area still lists epic **#4707** (closed) instead of the successor **#6943** that
  `issue_streams.yaml`'s `infra-harness` stream has carried since #6943 replaced it.
- No `open-model-data` area (epic #6321) at all.

Not fixing inline — touches `work_router.py`'s live area-tagging behavior for every work-board
consumer, which is its own change with its own blast radius, not a change to bundle into a
graph-map design doc. Filing so the drift is tracked and someone (or a generator script, if one
should exist) reconciles the two registries.

X-Agent: claude-monitor

krisztiankoos:
Closed via merged PR #7357 (3296806e96 on main, 2026-08-26).

Issue #7295, opened 2026-08-25
[monitor][ui] Epics & areas relationship map — visual view of streams → epics → issues with status/relations (#7177)


krisztiankoos:
Parent epic: #7177 (Monitor API + UI). Requested by operator 2026-08-25.

## What
A **visual relationship view** of the fleet's work structure — the operator wants to see, in one glance:
the unique **areas** and **epics**, their **relations**, their **status**, and their **issues**. Today's
`dashboards/epics.html` is a flat table of registered epics; this is the missing *graph/map* view.

## Data we already have (no new sources needed for v1)
- **Areas = registry streams**: `scripts/config/issue_streams.yaml` — each stream has a `title` and
  `epics: [numbers]` (e.g. `monitor` → [7177], `atlas-practice` → [4387,4700], `infra-harness` → [6943], …).
  This is the SSOT for "which area owns which epic."
- **Per-epic status**: `GET /api/epics/v1` — rows keyed by `stream_id` with `stream_name`, `title`,
  `lease` (active driver/host), `session_state`, `last_state`/`last_decision`/`last_next_action`,
  `registered`, `registry_status`.
- **Relations / issues**: GitHub issue graph — epic→sub-issue (native sub-issues + `#N` body refs),
  cross-references between issues, open/closed counts. `scripts/orchestration/issue_stream_audit.py`
  already walks stream membership.

## Shape (design owner refines; this is the WHAT)
1. **Backing endpoint** — a graph feed, e.g. `GET /api/epics/graph/v1`, assembling: nodes (areas, epics,
   optionally hot issues), edges (area→epic "contains", epic→epic parent/child, issue cross-refs), and
   per-epic rollups (open/closed issue counts, lease/status). Read-only, loopback like the rest; MUST
   pass the OPSEC sweep (no path/host leaks — it will be a NEW route in the sweep denominator).
2. **View** — a new `dashboards/epics-map.html` (or a tab on `epics.html`): areas as clusters, epics as
   nodes colored by status (active lease / released / expired / no-lease; open-issue heat), edges for
   relations, click-through to an epic detail panel with its issues + GitHub links. Self-contained
   (inline SVG/canvas — no external CDN), theme-aware, degrades if a feed field is missing.

## Acceptance
- One screen answers: what areas exist, which epics sit in each, how they relate, who's driving each,
  lease/issue status, and a link into the issues. Live from `/api/epics/graph/v1`.
[...]

PR #7312, opened 2026-08-25
feat(monitor): GET /api/epics/graph/v1 backing endpoint (#7295)

krisztiankoos:
Parent: #7295 (design doc merged as #7308). Implements the backing endpoint per
`docs/design/monitor-epics-graph-map.md` §4.1/§4.3/§5 — dashboard view (`dashboards/epics-map.html`)
is a separate follow-up PR gated on this one, per the design's sequencing.

**Note (driver):** this dispatch (agy) finished `status=done` without pushing or opening a PR —
same silent-exit pattern as the two cursor dispatches earlier this session (#7311), so it isn't
cursor-specific. Independently verified the actual commit content before pushing it myself:

- `tests/api/opsec_sweep/` — 19/19 passed.
- `tests/api/test_epics_router.py` (new) — 13/13 passed.
- Full `tests/api -n 4` — 368 passed, 1 skipped (pre-existing, unrelated).
- `ruff check` on all changed files — clean.

**What it does:**
- `GET /api/epics/graph/v1` in `epics_router.py`, response shape per design §4.1.
- Mirrors `GET /api/issues/streams`'s caching pattern exactly (read-cache / stale-fallback /
  schedule-refresh / off-event-loop) rather than any live `gh` call on the request path.
- Adds `open_issue_titles` to `classify()`'s cached report (design §4.3's flagged gap).
- Area layer from `issue_streams.yaml` only, not the stale `fleet_taxonomy.yaml` (#7306).
- OPSEC sweep: `FROZEN_HTTP_OPERATION_COUNT` 280→281, re-derived digest, isolated fixture entry
  for the new route — sweep passing is the proof the digest is correct.

Requesting cross-family review before merge.

X-Agent: claude-monitor

### DIFF
--- a/scripts/api/epics_router.py
+++ b/scripts/api/epics_router.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import base64
 import binascii
 import logging
@@ -12,7 +13,8 @@ from pathlib import Path
 from threading import Lock
 from typing import Any
 
-from fastapi import APIRouter, Request
+import yaml
+from fastapi import APIRouter, Query, Request
 from fastapi.responses import JSONResponse
 
 from agents_extensions.shared.session_streams.db import SessionStreamDatabase, default_database_path
@@ -43,12 +45,14 @@ from agents_extensions.shared.session_streams.store import (
 from scripts.api.config import LIVE_REPO_ROOT
 from scripts.api.observer_presence import _direct_loopback_peer
 from scripts.api.occupancy_sanitize import opaque_host_id, safe_field
+from scripts.orchestration import issue_stream_audit as audit
 from scripts.orchestration.thread_handoff import _bundle_extract, _bundle_secret_hits
 
 router = APIRouter()
 logger = logging.getLogger(__name__)
 
 SCHEMA = "remote-epic-lifecycle.v1"
+GRAPH_SCHEMA = "epics-graph.v1"
 DEFAULT_TTL_SECONDS = 15 * 60
 MAX_DIGEST_LIMIT = 100
 TOKEN_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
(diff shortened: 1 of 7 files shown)


krisztiankoos:
**Cross-family review: grok (grok-4.6) — REJECT.**

Independently re-verified both findings myself:

1. **[blocker] Privacy leak.** `open_issue_titles` (full open-issue title map) is added to the
   audit report at `issue_stream_audit.py:972` but is NOT in `PRIVATE_CACHE_KEYS`
   (`issue_stream_audit.py:59`, only has `effective_membership`/`open_issue_numbers`). Both
   `issues_router.py:251` and `sources_public.py` strip based on `PRIVATE_CACHE_KEYS` only —
   confirmed live. This means the full open-issue title map would leak through the EXISTING
   public `GET /api/issues/streams` endpoint on the next audit refresh. Design §4.3 asked for a
   cache-internal map for the graph endpoint's own use, not a new public field on an unrelated
   existing endpoint.
2. **Vacuous test.** `test_epics_graph_store_lease_and_decision_passthrough` guards its real
   assertions behind `if "epic:7178" in epics_by_id`. `7178` is a lease-lifecycle test fixture ID
   used elsewhere in the same file — confirmed it does not appear anywhere in
   `scripts/config/issue_streams.yaml` (monitor's only epic is 7177). Graph nodes come from that
   yaml, so the guard never fires and the assertions never run; pytest reports a pass either way.

Genuinely good catch, especially #1. Sent back to the authoring lane for a fix — not fixing
inline.

X-Agent: claude-monitor

krisztiankoos:
Fix pushed (\`4632e9c785\`) addressing both REJECT findings. Independently re-verified myself
before requesting re-review (dispatch again settled \`done\` without pushing — same #7311 gap,
now 4/4 write-capable dispatches this session, pushed manually):

- \`open_issue_titles\` added to \`PRIVATE_CACHE_KEYS\`; extended strip tests in
  \`tests/test_research_registry_p4.py\`, \`tests/test_work_api.py\`, \`tests/test_issue_stream_audit.py\`.
- Passthrough test now claims the real registered \`epic:7177\` (was fabricated \`epic:7178\`,
  absent from \`issue_streams.yaml\`); the \`if\` guard is gone, assertions run unconditionally.
- Ran myself: \`opsec_sweep\` + \`test_epics_router.py\` + \`test_issue_stream_audit.py\` — 84 passed.
  \`test_research_registry_p4.py\` + \`test_work_api.py\` — 89 passed, 1 environment-only failure
  (\`.venv\` absent in dispatch worktree, the known #6858 class — confirmed passes from primary
  checkout). Full \`tests/api -n 4\` — 367 passed, 1 skipped, 1 pre-existing flake (live-uvicorn
  \`test_release_snapshot\`, reproduced identically against unmodified \`main\`). \`ruff\` clean.

Requesting re-review from the original reviewer (grok) to confirm both findings are closed.

X-Agent: claude-monitor

krisztiankoos:
**Re-review: grok — APPROVE.** Both findings confirmed fixed, with a genuinely thorough
re-verification: a synthetic canary-title leak proof (planted \`LEAK-CANARY-TITLE-SHOULD-NOT-APPEAR\`,
confirmed absent from \`GET /api/issues/streams\`'s actual response body), and a load-bearing check
on the passthrough test showing it would now actually FAIL if the lease/state/decision wiring
broke (not just pass vacuously).

Merging.

X-Agent: claude-monitor

PR #7337, opened 2026-08-25
refactor(monitor): migrate session-streams cluster to MonitorContext (#7269 step 1)

krisztiankoos:
Parent: #7269 step 1 (#7319). Design: docs/design/monitor-api-app-factory.md §4.1-4.3, §6-7.
First-ever router family migration to MonitorContext — establishes the pattern for the 16
remaining steps under #7269.

**Note (driver):** this dispatch (agy) settled done without pushing — same pattern as 4 earlier
dispatches this session (now fixed upstream as #7311/#7315, merged during this dispatch's run;
my local delegate.py was pre-fix when I launched it). Independently verified before pushing:

- tests/api/opsec_sweep/ + tests/api/test_app_factory.py + the three touched API test files —
  62 passed.
- Full tests/api -n 4 — 368 passed, 1 skipped, 1 pre-existing flake (live-uvicorn
  test_release_snapshot, previously confirmed identical on unmodified main).
- ruff clean.

**What it does:**
- Migrates session_streams_router.py, rollover_router.py, session_router.py, rules_router.py to
  Depends(get_ctx) instead of module globals.
- Removes session_streams_router.py from DB_ACCESS_ALLOWLIST (21→20) — it no longer calls
  SessionStreamDatabase directly.
- Drops the 12 now-redundant OPSEC sweep seams this family owned (198→186 unique logical seams).
- Adds a fixture-isolation test for this family (design §7 point 4 — the first such test, since
  step 0 explicitly couldn't write one).
- **Real cross-file fix caught along the way**: occupancy_local.py and fleet_workers_collect.py
  (step 3's cluster) were importing session_streams_router.py's now-removed private helpers
  (_db_path, _repo_root) directly. Relocated the equivalent logic into occupancy_local.py itself
  rather than leaving a broken import — a real coupling the design doc's per-step-independence
  framing didn't anticipate; worth flagging for the reviewer and for whoever picks up step 3.

Requesting cross-family review before merge.

X-Agent: claude-monitor

### DIFF
--- a/docs/design/count_opsec_fixture_seams.py
+++ b/docs/design/count_opsec_fixture_seams.py
@@ -24,7 +24,6 @@ from scripts.api import (
     docs_router,
     entire_context_router,
     epics_router,
-    fleet_router,
     git_hygiene_router,
     governance_router,
     images_router,
@@ -32,7 +31,6 @@ from scripts.api import (
     project_state_collect,
     project_state_router,
     repository_authority,
-    session_streams_router,
     site_router,
     state_helpers,
     work_router,
@@ -167,21 +165,6 @@ def replay_isolated_fixture(monkeypatch: MonkeypatchRecorder, root: Path) -> Non
     monkeypatch.setattr(entire_context_router, "load_provider_capabilities", lambda _root: {})
     monkeypatch.setattr(reap_worktrees, "_run", lambda *_args, **_kwargs: (0, "", ""))
     monkeypatch.setattr(atlas_job, "primary_checkout_root", lambda: root)
-    monkeypatch.setattr(session_streams_router, "_repo_root", lambda: root)
-    monkeypatch.setattr(
-        session_streams_router,
-        "_db_path",
-        lambda: root / "stores" / "session-streams.sqlite3",
-    )
-    monkeypatch.setattr(session_streams_router, "_store", lambda: _FixtureSessionStore())
-    monkeypatch.setattr(session_streams_router, "list_handoff_candidates", lambda _root: [])
-    monkeypatch.setattr(
-        session_streams_router,
-        "diagnose_handoff",
-        lambda _store, stream_id: _FixtureHandoff(stream_id),
-    )
-    monkeypatch.setattr(session_streams_router, "list_projection_receipts", lambda *_a, **_k: [])
-    monkeypatch.setattr(session_streams_router, "detect_projection_drift", lambda *_a, **_k: {})
     monkeypatch.setattr(worktree_containment, "primary_checkout_dirty_status", lambda _s: {})
     monkeypatch.setattr(cold_start_board, "_get_local_git_info", lambda: {})
     monkeypatch.setattr(
--- a/scripts/api/comms_router.py
+++ b/scripts/api/comms_router.py
@@ -1046,6 +1046,9 @@ def _scan_live_activity(minutes: int = 15) -> list[dict]:
     activities = []
 
     # 1. Scan all orchestration state files modified recently
+    if not CURRICULUM_ROOT.is_dir():
+        return activities
+
     for track_dir in CURRICULUM_ROOT.iterdir():
         if not track_dir.is_dir():
             continue
@@ -1127,6 +1130,9 @@ def _scan_recent_completions(minutes: int = 60) -> list[dict]:
     now = time.time()
     cutoff = now - (minutes * 60)
     completions = []
+
+    if not CURRICULUM_ROOT.is_dir():
+        return completions
 
     for track_dir in CURRICULUM_ROOT.iterdir():
         if not track_dir.is_dir():
--- a/scripts/api/fleet_workers_collect.py
+++ b/scripts/api/fleet_workers_collect.py
@@ -33,6 +33,7 @@ from scripts.api.occupancy_local import (
     read_session_streams,
     resolve_launcher_host_id,
     self_host_opaque_ids,
+    session_streams_db_path,
 )
 from scripts.api.occupancy_sanitize import CLOUD_OBSERVER_HOST_ID, opaque_host_id, safe_field
 from scripts.api.project_state_store import (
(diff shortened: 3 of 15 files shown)


krisztiankoos:
**Cross-family review: grok (grok-4.6) — REJECT.**

Independently re-verified the blocking finding myself:
`agents_extensions/shared/session_streams/db.py:129` raises `SessionStreamDatabaseError(RuntimeError)`
on a missing DB file — NOT `FileNotFoundError`. The migrated router's `except (NotFoundError,
FileNotFoundError)` catch clause (session_streams_router.py:74/95) no longer catches it, so a
missing-DB request now falls through to FastAPI's global handler → 500, where pre-migration it was
a clean 404. Confirmed: the `if store is None:` guards (lines 70/91/151) are dead code —
`MonitorContext._stores()` always constructs a `SessionStreamStore`, never `None`, regardless of
whether the underlying file exists.

This is exactly the design doc's "zero behavior change" requirement being violated — a real,
verified regression, not review noise. grok also did a live probe against `fixture_context` with
no DB file confirming the actual status codes (500 vs 404) — very thorough work, genuinely good
catch. Findings 4-7 (allowlist, seam deletions, occupancy relocation, session/rules/rollover
routes) are all confirmed correct and don't need re-litigating.

Sent back to the authoring lane for a fix.

X-Agent: claude-monitor

krisztiankoos:
Fix pushed (\`232c8280ca\`) — Option A from the fix brief (restore the file-existence check via
\`ctx.roots.session_streams_db_path.is_file()\`, remove the dead \`if store is None:\` guards,
align \`/v1/drift\` to the same 404 contract). Independently re-verified myself:

- Reproduced grok's exact probe (\`create_app(fixture_context(tmp_path))\`, no DB file) myself:
  both \`/v1/status/epic:4707\` and \`/v1/digest/epic:4707\` now return **404** again, not 500.
- \`opsec_sweep\` + \`test_app_factory.py\` + the three touched API test files — 63 passed.
- Full \`tests/api -n 4\` — 368 passed, 1 skipped, the same known pre-existing flake. \`ruff\` clean.
- A permanent regression test for this exact case is now in \`tests/test_session_streams_api.py\`.

Requesting re-review from the original reviewer (grok) to confirm the fix closes the finding.

X-Agent: claude-monitor

krisztiankoos:
**Re-review: grok — APPROVE.** Reproduced the exact same live probe: \`/v1/status\` and
\`/v1/digest\` both 404 now (were 500), \`/v1/drift\` also fixed (not left as a follow-up — the
commit brought it in-scope). Re-ran both test suites, same results as before (same pre-existing
unrelated flake). No new findings in the fix commit.

Merging.

X-Agent: claude-monitor

krisztiankoos:
CI's \`Pytest fastlane\` failed on a real, unrelated house lint
(\`test_repo_test_suite_is_clean\` — forbids a hardcoded epic-ID literal in a test assertion
without an \`# allow-hardcoded-epic:\` annotation). The new regression test's \`epic:4707\`
assertions were missing it. Fixed (\`d0025d8831\`) — two-line annotation only, no logic change.
Reproduced the exact failing test locally, now passes, plus the two other affected suites.

X-Agent: claude-monitor

krisztiankoos:
Fix pushed (\`1320b745dc\`) for the merge-queue kick. Migrated the 3 monkeypatch sites in
\`tests/test_monitor_client_sdk.py\` to swap \`api_main.app.state.ctx\` to a
\`fixture_context(...)\`-built context instead of monkeypatching the now-removed
\`session_router.PROJECT_ROOT\`/\`rules_router.PROJECT_ROOT\` module attributes — the correct fix
per the design's own anti-pattern-elimination goal, not resurrecting the dead constant.

Also found (proactively, while making the SDK test fixture-context-aware) two zero-production-risk
defensive gaps this exposed: \`state_router.py\`'s \`manifest()\` route now explicitly passes
\`ctx.roots.project_root\` to \`rules_hash()\`/\`session_hash()\` instead of relying on their
internal default, and \`comms_router.py\`'s \`_scan_live_activity\`/\`_scan_recent_completions\`
now guard against a nonexistent \`CURRICULUM_ROOT\` (which \`fixture_context\` doesn't create on
disk unless a test explicitly does). Both changes are behavior-identical for \`production_context()\`
— they only change what happens when the directory genuinely doesn't exist, which never happens
in production.

Independently verified before pushing:
- \`tests/test_monitor_client_sdk.py\` + \`opsec_sweep\` + \`test_app_factory.py\` +
  \`test_manifest_api.py\` — 67 passed.
- Grepped the WHOLE repo for any other consumer of the removed constants — none found.
- Ran the full blast-radius test set (49 files across the repo that import the affected modules,
  not just \`tests/api/\`) against BOTH the dispatch worktree and primary checkout to separate real
  regressions from sparse-worktree artifacts: 18 of 21 worktree failures don't reproduce on the
  full-data primary checkout (missing curriculum/wiki — sparse-checkout artifact, not code); the
  remaining 3 are a pre-existing, unrelated local-environment issue (\`PlaneRootAnchorError\` —
  this Mac's local fleet-comms plane is retired) that also reproduces identically on unmodified
  \`main\`. Zero regressions from this fix.
- \`ruff\` clean.

Re-arming the merge queue.

X-Agent: claude-monitor

krisztiankoos:
**Third review pass: grok — APPROVE.** Confirmed with tool-backed proofs: the ctx-swap correctly
isolates the shared TestClient singleton (monkeypatch teardown restores production ctx), and
\`rules_hash()\`/\`session_hash()\` produce byte-identical hashes whether called with the implicit
default or the explicit \`ctx.roots.project_root\` (measured directly: \`ctx.roots.project_root ==
Path(config.PROJECT_ROOT)\` is True). The comms_router guards are confirmed production-identical
(only change behavior when the directory is literally absent). Also independently hit the same
sparse-worktree \`CURRICULUM_ROOT\` absence I found, confirming it as the reason the guards are
necessary for the test suite, not a production risk.

Re-arming the merge queue.

X-Agent: claude-monitor

